PR #4, opened 2025-01-09
Update day16.rs

dfarnham:
This isn't a proposed final fix, just an attempt to show the overflow error my input generates.

This is repository is absolutely spectacular!

Please feel free to reach out for more details if needed.

### DIFF
--- a/src/year2024/day16.rs
+++ b/src/year2024/day16.rs
@@ -89,10 +89,36 @@ pub fn parse(input: &str) -> Input {
         // Reverse direction and subtract cost.
         let left = (direction + 3) % 4;
         let right = (direction + 1) % 4;
+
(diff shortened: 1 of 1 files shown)


maneatingape:
Great catch! Overlooked this use case.

I'd like to make sure you get a contributor credit by merging this PR but it needs to be cleaned up first.

How about simplifying by doing the following:
* Change all instances `u32` to `i32` e.g. 
```
    let mut seen = grid.same_size_with([i32::MAX; 4]);
    let mut lowest = i32::MAX;
```

This should fix the bug and be equally performant to the original.
Can you check this fixes the issue?

maneatingape:
Fixed in PR #5

PR #6, opened 2025-01-14
Year 2022 Day 20 regression

dfarnham:


maneatingape:
Thanks for the bug finding!

Does this change fix the issue? I'm not sure how it would produce a different result.

dfarnham:
No, sorry for the confusion.  Factoring out the multiplication just saves an instruction is all.  If you navigate to my AOC repo you can test my puzzle input across your previous tree algorithm which produces the correct answer.

maneatingape:
[Fixed](https://github.com/maneatingape/advent-of-code-rust/commit/076ea1a2f178c7eab64b2ac6936b8ec3ae30dec7).

Please verify against your input.

dfarnham:
Verified fixed on my input.

maneatingape:
Thanks for flagging the issue!

PR #7, opened 2025-01-15
Remove check in day 13

will-lynas:
I don't think this check is needed. (For my input at least) I still get the right answer without it.

image
I think the above comment about 100 presses is just to give people some bounds to allow them to try a brute force solution.


maneatingape (approved):

maneatingape:
Nice simplification!

PR #11, opened 2025-07-12
perf: improve performance of 2018 day 2 part 2

Rolv-Apneseth:
Hi again. This is a quick one for improving the performance of 2018 day 2 part 2 (85μs => 15μs for me). Let me know if you want any changes, and no hard feelings if you don't want the PR at all.

### DIFF
--- a/src/year2018/day02.rs
+++ b/src/year2018/day02.rs
@@ -65,7 +65,8 @@ pub fn part2(input: &[&[u8]]) -> String {
                 return id1
                     .iter()
                     .zip(id2)
-                    .filter_map(|(a, b)| (a == b).then(|| char::from(*a)))
+                    .filter(|&(a, b)| a == b)
+                    .map(|(a, _)| char::from(*a))
                     .collect();
             }
         }
--- a/src/year2018/day02.rs
+++ b/src/year2018/day02.rs
@@ -1,5 +1,4 @@
 //! # Inventory Management System
-use crate::util::hash::*;
 
 pub fn parse(input: &str) -> Vec<&[u8]> {
     input.lines().map(str::as_bytes).collect()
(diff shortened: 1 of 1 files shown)


src/year2018/day02.rs
```diff
@@ -44,29 +43,33 @@ pub fn part1(input: &[&[u8]]) -> u32 {
...
-    let mut seen = FastSet::with_capacity(input.len());
-    let mut buffer = [0; 32];
-
-    // Use a set to check for duplicates after replacing a single character with '*' in each column.
-    for column in 0..width {
-        for &id in input {
-            buffer[0..width].copy_from_slice(id);
-            buffer[column] = b'*';
+            let mut diff = false;
+            for (a, b) in id1.iter().zip(id2) {
```
maneatingape, line 53: This causes a performance regression on my input, 87 µs to 129 µs.

Two nested loops have woprst case quadratic complexity `O(n²)`. 
Hashing is a constant time operation so the original solution worst case complexity is `O(1) * 26 * O(n) = O(n)`.

So it looks like you're getting lucky, while I'm getting unlucky with the input order.

maneatingape (changes requested):

src/year2018/day02.rs
Rolv-Apneseth: I am no expert in `O(N)` notation, but wouldn't the `N` here be the number of IDs, so it should be lower than `O(N^2)` (second loop starts at the index of the first loop)?

Anyway, fair enough about the performance being worse. Just luck based that it's so much faster for my input then. I'll close the PR if that's alright with you, but out of curiosity would mind sharing your input too?

maneatingape: > I am no expert in O(N) notation, but wouldn't the N here be the number of IDs, so it should be lower than O(N^2) (second loop starts at the index of the first loop)?

Say you have 10 ids. First inner loop iteration will check 9 items, next iteration 8 and so on...
9 + 8 + 7 + ... + 2 + 1 = [n(n + 1) / 2](https://en.wikipedia.org/wiki/1_%2B_2_%2B_3_%2B_4_%2B_%E2%8B%AF) = (1/2)(n² + n)
So you save half the comparisons but it's still overall quadratic.

> Anyway, fair enough about the performance being worse. Just luck based that it's so much faster for my input then. I'll close the PR if that's alright with you, but out of curiosity would mind sharing your input too?

If you re-order the elements (for example swap the first and second halves) you can change the performance drastically. With a few random swaps I got benchmarks from 33 µs to 250 µs.

maneatingape: I still think it might be possible to improve the performance by taking advantage of the special structure of the input (always 26 lower case ASCII characters).

maneatingape:
BTW I welcome PRs (even if they don't always work out). It's always an interesting discussion and could spark an improvement either in this or another solution.

Rolv-Apneseth:
If you're interested @maneatingape, it was just a simple change but for me it's giving around 30-45% speed up:

```rust
pub fn part2(input: &[&[u8]]) -> String {
    let width = input[0].len();

    let mut seen = FastSet::with_capacity(input.len());

    // Use a set to check for duplicates by comparing the prefix and suffix of IDs excluding one
    // column at a time.
    for column in 0..width {
        for &id in input {
            let prefix = &id[..column];
            let suffix = &id[column + 1..];

            if !seen.insert([prefix, suffix]) {
                // Convert to String
                return prefix.iter().chain(suffix).cloned().map(char::from).collect();
            }
        }

        seen.clear();
    }

    unreachable!()
}
```

maneatingape (approved):
Neat! Faster and more elegant.

79 µs => 49 µs.

Rolv-Apneseth:
Missed that lint - not getting these from clippy locally but yeah `cloned` -> `copied` makes sense

PR #30, opened 2026-01-30
Speed up 2025 day 9 part 1 max rectangle computation

ebblake:
## Description

Compute the max rectangle in O(n log n) effort (643 calls to .max) rather than O(n^2) (6206 calls).  Do this by recognizing that you can divide and conquer the pairing: for the mid-point of the upper left region, you can detemine a corresponding pivot point in the lower right region, and of the four quadrants resulting from points remaining on either side of your pivots, only two of the four combinations can produce a larger rectangle.

Cuts runtime of part 1 on my machine from 11us to 4us.

Inspired by https://codeforces.com/blog/entry/128350

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

  specific intrinsics.
- [x] Tests pass

      cargo test

- [x] Code is formatted

- [x] Code is linted

### DIFF
--- a/src/year2025/day09.rs
+++ b/src/year2025/day09.rs
@@ -50,8 +50,8 @@ pub fn part1(tiles: &[Tile]) -> u64 {
     let (bottom_left_tiles, bottom_right_tiles) =
         get_potential_left_corner_tiles(tiles.iter().copied().rev());
 
-    find_largest_from_all_corners(&top_left_tiles, &bottom_right_tiles)
-        .max(find_largest_from_all_corners(&bottom_left_tiles, &top_right_tiles))
+    find_largest_from_all_corners(&top_left_tiles, &bottom_right_tiles, true)
+        .max(find_largest_from_all_corners(&bottom_left_tiles, &top_right_tiles, false))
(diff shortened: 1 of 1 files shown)


ebblake:
Any thoughts on accepting this part 1 speedup, or at least revising the patch to document how a bit more code can lead to faster part 1 runtimes?  The patch drops from O(n^2) to O(n log n).  On my machine, I'm seeing part1() drop from 11us to 4us; which is about a 10% speedup on the overall day.

maneatingape:
> Any thoughts on accepting this part 1 speedup, or at least revising the patch to document how a bit more code can lead to faster part 1 runtimes? The patch drops from O(n^2) to O(n log n). On my machine, I'm seeing part1() drop from 11us to 4us; which is about a 10% speedup on the overall day.

This delay in reviewing is mostly on me, the current solution is a submission from another contributor and I haven't got around to throughly understanding it yet.

That said this is a reasonable speed up. It's a little more complex than I'd prefer but the tradeoff is acceptable.

PR #34, opened 2026-02-02
Track revisited points by segment intersection

ebblake:
## Description

Tracking line segments and searching for intersections is faster than hashing every point along the way.  On my machine, cuts part2 runtime from 3.5us down to 0.6us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

  specific intrinsics.
- [X] Tests pass

      cargo test

- [X] Code is formatted

- [X] Code is linted

### DIFF
--- a/src/year2016/day01.rs
+++ b/src/year2016/day01.rs
@@ -1,17 +1,45 @@
 //! # No Time for a Taxicab
 //!
-//! The solution is short as it leverages three utility classes, [`hash`] for speedy sets,
+//! The solution is short as it leverages two utility classes,
 //! [`parse`] for extracting integers from surrounding text and [`point`] for two dimensional
 //! rotations and translations.
 //!
-//! [`hash`]: crate::util::hash
+//! For part two, it is faster to remember line segments and check for intersections than
+//! to store information on every intermediate point visited.
+//!
 //! [`parse`]: crate::util::parse
 //! [`point`]: crate::util::point
-use crate::util::hash::*;
 use crate::util::parse::*;
 use crate::util::point::*;
 
 type Pair = (u8, i32);
+
+// Represent the line segment between two points.
+struct Segment {
+    x1: i32,
+    x2: i32,
+    y1: i32,
+    y2: i32,
+}
+
+fn minmax(a: i32, b: i32) -> (i32, i32) {
+    if a < b { (a, b) } else { (b, a) }
+}
+
+impl Segment {
+    fn new(first: Point, second: Point) -> Self {
+        let (x1, x2) = minmax(first.x, second.x);
+        let (y1, y2) = minmax(first.y, second.y);
+        Segment { x1, x2, y1, y2 }
+    }
+
+    // Return the point of intersection between two orthogonal segments, if there is one.
+    fn intersects(&self, other: &Segment) -> Option<Point> {
+        let overlap =
+            !(other.x2 < self.x1 || other.x1 > self.x2 || other.y2 < self.y1 || other.y1 > self.y2);
+        overlap.then_some(Point::new(self.x1.max(other.x1), self.y1.max(other.y1)))
+    }
+}
 
 pub fn parse(input: &str) -> Vec<Pair> {
     let first = input.bytes().filter(u8::is_ascii_uppercase);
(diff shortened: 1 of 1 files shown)


ebblake:
Once again, it probably shows that I'm not very familiar with the best idioms for passing references around; I welcome advice on how to clean this up.  This particular day's puzzle is not in the overall hotspot, but I took a shot at it anyways thanks to https://www.reddit.com/r/adventofcode/comments/1qsousl/2016_day_12_in_review_no_time_for_a_taxicab/

src/year2016/day01.rs
```diff
@@ -1,18 +1,60 @@
...
+    // Whether dim0 represents x or y depends on context, but doesn't matter to the final Manhattan
+    // distance; lo and hi represent the points past first on the way to and including second.
+    #[inline]
+    #[must_use]
+    const fn new(first: Point, second: Point) -> Self {
+        if first.x == second.x {
+            let lo = if first.y < second.y { first.y + 1 } else { second.y };
+            let hi = if first.y < second.y { second.y } else { first.y - 1 };
+            Segment { dim0: first.x, lo, hi }
+        } else {
```
ebblake: The commit plays games with the unusual (first,second] layout (rather than both ends inclusive or a more common [first,second) half-open interval) so that a brand new segment does not conflict with the most-recently added orthogonal segment.  I also tried this approach, but it was slightly slower:
```
diff --git i/src/year2016/day01.rs w/src/year2016/day01.rs
index 675b22b..0878f06 100644
--- i/src/year2016/day01.rs
+++ w/src/year2016/day01.rs
@@ -24,17 +24,17 @@ struct Segment {

 impl Segment {
     // Whether dim0 represents x or y depends on context, but doesn't matter to the final Manhattan
-    // distance; lo and hi represent the points past first on the way to and including second.
+    // distance; lo and hi represent the points from first to second, inclusive.
     #[inline]
     #[must_use]
     const fn new(first: Point, second: Point) -> Self {
         if first.x == second.x {
-            let lo = if first.y < second.y { first.y + 1 } else { second.y };
-            let hi = if first.y < second.y { second.y } else { first.y - 1 };
+            let lo = if first.y < second.y { first.y } else { second.y };
+            let hi = if first.y < second.y { second.y } else { first.y };
             Segment { dim0: first.x, lo, hi }
         } else {
-            let lo = if first.x < second.x { first.x + 1 } else { second.x };
-            let hi = if first.x < second.x { second.x } else { first.x - 1 };
+            let lo = if first.x < second.x { first.x } else { second.x };
+            let hi = if first.x < second.x { second.x } else { first.x };
             Segment { dim0: first.y, lo, hi }
         }
     }
@@ -88,7 +88,7 @@ pub fn part2(input: &[Pair]) -> i32 {
         let target = position + direction * amount;
         let segment = Segment::new(position, target);

-        for other in &mut *curr {
+        for other in curr.iter().take(curr.len().saturating_sub(1)) {
             if let Some(point) = segment.intersects(&other) {
                 return point.manhattan(ORIGIN);
             }
```

src/year2016/day01.rs
```diff
@@ -1,18 +1,60 @@
...
+            let lo = if first.x < second.x { first.x + 1 } else { second.x };
+            let hi = if first.x < second.x { second.x } else { first.x - 1 };
```
ebblake: Too bad std::cmp::minmax is not yet stable.  It would have been nice to use here. https://doc.rust-lang.org/std/cmp/fn.minmax.html

src/year2016/day01.rs
```diff
@@ -35,18 +77,24 @@ pub fn part1(input: &[Pair]) -> i32 {
...
+    // Store two lists of segments, one for each orthogonal direction.
+    let mut east_west = Vec::<Segment>::with_capacity(input.len() / 2);
+    let mut north_south = Vec::<Segment>::with_capacity(input.len() / 2);
+    let (mut curr, mut next) = (&mut east_west, &mut north_south);
 
     for &(turn, amount) in input {
         direction =
             if turn == b'L' { direction.counter_clockwise() } else { direction.clockwise() };
+        let target = position + direction * amount;
+        let segment = Segment::new(position, target);
```
maneatingape: Prevent overlapping segment is `let segment = Segment::new(position + direction, target);`

src/year2016/day01.rs
```diff
@@ -1,18 +1,60 @@
...
+            Segment { dim0: first.x, lo, hi }
+        } else {
+            let lo = if first.x < second.x { first.x + 1 } else { second.x };
+            let hi = if first.x < second.x { second.x } else { first.x - 1 };
+            Segment { dim0: first.y, lo, hi }
+        }
+    }
+
+    // Return the point of intersection between two orthogonal segments, if there is one.
+    #[inline]
```
maneatingape: Remove uneccessary attributes. The compiler does a good job of automatically inlining, so `#[inline]` is only needed when it guesses wrong and benchmarking shows an improvement. Interestingly sometimes `#[inline]` is slightly *slower*, so always check.

maneatingape (changes requested):

src/year2016/day01.rs
ebblake: I'm not seeing obvious differences whether I use or drop `#[inline]` on either `new` or `intersects`, so I left it off.

ebblake:
I just noticed that 2019 day 3 does intersection tracking more efficiently with a Btree to limit to plausible ranges. It is also possible to do range limiting via a pair of stacks rather than a Btree for the same amount of comparisons, but different overhead on how the ranges are tracked.

PR #64, opened 2026-04-21
Force tighter array bounds during search

ebblake:
## Description

The part 2 search intentionally allows searching beyond the tight rectangle of the target (in fact, the problem statement mentions that an oversized grid will be necessary to find the optimal route). However, the target is so far off in the y direction (which we picked as width) that the frontier tends to explore fairly far into the x direction (which we picked as height), at least for lower y values. In practice, any jaunt 20 steps past the target to find a better vein of travel that avoids tool changes that would be present on a path with less of a detour must be balanced by another 20 steps of return, and in practice the grid tends to not have enough continuity for that jaunt to actually pan out.  Manually constraining the bounds to fit within +30 results in about one quarter of the grid size needing population, without penalizing the actual result (in my input, the correct answer was still reached even with a cap of +20).

On my machine, this speeds up runtime from 4.3ms to 1.5ms.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

  specific intrinsics.
- [X] Tests pass `cargo test`


ebblake:
I'm trying to determine if a smarter A* heuristic could help without my hand-wavy hard clamp on y bounds. What if, while computing the erosion values for the grid along increasing x, you kept a running counter of the run lengths of tiles that do not require a tool change for that line.  If we could have a reasonably statistical confidence that a change in (say) 50 x is likely to cross a tool change boundary 3 times (made up values for discussion) then that cost could possibly be added into the A* heuristic to make low x high y points more expensive (more likely tool changes from the higher delta x to the goal) than high x low y (progress towards the goal already includes actual tool changes, with fewer x remaining to cause more), and thus favor the frontier that moves along a narrower corridor to the goal than when just Manhattan distance would otherwise treat them as equals. But it might risk being inadmissible if we overestimate the average run length.

maneatingape:
The existing grid width limits are also hand-wavy, so this is a better [SWAG](https://en.wikipedia.org/wiki/Scientific_wild-ass_guess) 😄 

I get a similar speed up, 3ms -> 1.1ms.

src/year2018/day22.rs
```diff
@@ -96,7 +96,11 @@ pub fn part2(input: &Input) -> i32 {
             for next in ORTHOGONAL.map(|o| point + o) {
                 // We don't need an additional check that the tool cannot be used in the
                 // destination region, as the time check will take care of that.
-                if next.x >= 0 && next.y >= 0 && time + 1 < cave[next].minutes[tool] {
+                if next.x >= 0
```
maneatingape: Bounds check simplify to `cave.contains(next)`

src/year2018/day22.rs
```diff
@@ -76,7 +76,7 @@ pub fn part2(input: &Input) -> i32 {
     let mut todo: [_; BUCKETS] = from_fn(|_| Vec::with_capacity(1_000));
 
     // Add extra width and height so the search does not exceed the bounds of the grid.
-    let mut cave = scan_cave(input, width + 10, height + 140);
+    let mut cave = scan_cave(input, width + 10, height + 30);
```
maneatingape: Let's call these out explicitly by extracting to a `const` and mentioning in the header comment e.g. something like:

```
const WIDTH_BUFFER_SWAG = 10;
const HEIGHT_BUFFER_SWAG = 30;
```
"estimate" or some other synonym for "guess" will also work.

I should have done this when writing the first time! 😅

Optionally, we could collapse into just a single buffer of 30 e.g. 

```
const BUFFER_SWAG = 30;
...
let mut cave = scan_cave(input, width + BUFFER_SWAG, height + BUFFER_SWAG);
```

maneatingape (commented):
Nice speedup! 🔥

ebblake:
I also wonder if even more speed can be gained by having parse() populate the cave with the SWAG bounds, rather than part1 and part2 both populating caves of slightly different sizes but where the majority of cave2 population is repeating work.

maneatingape:
> I also wonder if even more speed can be gained by having parse() populate the cave with the SWAG bounds, rather than part1 and part2 both populating caves of slightly different sizes but where the majority of cave2 population is repeating work.

Yup, a smaller but potentially still worthy gain. 

We may have to expand the buffer, from the Reddit discussion thread [+31](https://www.reddit.com/r/adventofcode/comments/1sshn5l/comment/ohm1l75/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) and [+63](https://www.reddit.com/r/adventofcode/comments/1sshn5l/comment/ohm1l75/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

ebblake:
> We may have to expand the buffer, from the Reddit discussion thread +31 and +63

Yeah, I saw that. I'm now in the process of scraping the megathread to collect a pile of actual input triples, to then see what slop bounds are sufficient.  From the context, it wasn't clear if that was +63 from the axis (more like a slop of 53 if x=10) or +63 from the target (more like x=73 being reached on the winning path, if the target is x=10).  Hopefully one of the leaked megathread triples matches the reddit thread description.

ebblake:
Patch updated to explore the idea of sharing the part1/2 grid generation in parse().  Not sure I like it - pre-patch parse was sub-1ns and part1 was 40ns; post-patch parse is 115ns (but better than the 500ns with unchanged array bounds for part 2) while part1 is 1ns.  And that's with a SLOP that worked for me, but still not tested to work with other megathread inputs.

I do like the SLOP constant, but may rip out the first commit that tries to refactor things into parse() and just focus on the second patch.

src/year2018/day22.rs
```diff
@@ -34,58 +42,97 @@ use std::array::from_fn;
...
+            if point == target {
+                row[x] = depth % 20183;
+            } else {
+                row[x] = (prev * row[x] + depth) % 20183;
+            }
+            prev = row[x];
+            grid[point] = (row[x] % 3) as u8;
+        }
+    }
+
```
maneatingape, line 96: With tidy above insert this:

```
    let target = Point::new(target_col, target_row);
    grid[target] = ((depth % 20183) % 3) as u8;
```

src/year2018/day22.rs
```diff
@@ -34,58 +42,97 @@ use std::array::from_fn;
...
+    for y in 1..target_row + SLOP {
+        let mut prev = (16807 * y + depth) % 20183;
+        row[0] = prev;
+        grid[Point::new(0, y)] = (row[0] % 3) as u8;
+
+        for x in 1..target_col + SLOP {
+            let point = Point::new(x, y);
+            let x = x as usize;
+            if point == target {
```
maneatingape: This can simplify to just:

```
            row[x] = (prev * row[x] + depth) % 20183;
```

with code insertion below. (Fixing my own code 😆 )

src/year2018/day22.rs
```diff
@@ -34,58 +42,97 @@ use std::array::from_fn;
...
 }
 
-/// Build the minimum grid to the target then calculate the risk level.
+/// Calculate the risk level of the relevant subset of the overall cave grid.
 pub fn part1(input: &Input) -> i32 {
-    let [_, height, width] = *input;
-    let cave = scan_cave(input, width + 1, height + 1);
-    cave.bytes.iter().map(|r| r.erosion % 3).sum()
+    let Input { cave, height, width } = input;
+    cave.bytes
```
maneatingape, line 103: ❤️

src/year2018/day22.rs
```diff
@@ -34,58 +42,97 @@ use std::array::from_fn;
...
+    // always be greater than zero.
+    let mut cave = Grid::new(erosion.width, erosion.height, [i32::MAX; 3]);
+    for (i, level) in erosion.bytes.iter().enumerate() {
+        cave.bytes[i][*level as usize] = 0;
+    }
 
     // Start at origin with the torch equipped.
     todo[0].push((ORIGIN, TORCH));
-    cave[ORIGIN].minutes[TORCH] = 0;
+    cave[ORIGIN][TORCH] = 0;
```
maneatingape, line 131: Nice simplification

maneatingape (commented):
Code looks good.

Question on the base SLOP assumption?
Found one input on the megathread that breaks:
```
depth: 4845
target: 6,770
```
but I'm not sure if it's an actual input or theoretical?

IIUC u/musifter confimed that their input was +63 on the shortest path?
Perhaps it's worth DMing them and checking the actual input to confirm.

ebblake:
> Code looks good.
> 
> Question on the base SLOP assumption? Found one input on the megathread that breaks:
> 
> ```
> depth: 4845
> target: 6,770
> ```
> 
> but I'm not sure if it's an actual input or theoretical?

It looked real to me, since I also found it; another one in the megathread that caused issues was 6969,9,796.

> 
> IIUC u/musifter confimed that their input was +63 on the shortest path? Perhaps it's worth DMing them and checking the actual input to confirm.

I did not find a reproducer for that in the megathread, so yeah, a DM check is the next step.

A slop of 64 is not quite as much of a win as a slop of 30, but should still be faster than a slop of 140. I'm still investigating if there's other things that can be done to prune states for less work (I'm a bit annoyed that my m4 solution had to slow down from 3.0 to 4.5s to be generic to some of these wide-slop inputs from the megathread).

src/year2018/day22.rs
```diff
@@ -8,6 +8,14 @@
 //! Interestingly, benchmarking showed that adding the time to switch tools if we don't have the
 //! torch to the heuristic slowed things down.
 //!
+//! The various input files all have a target with a first coordinate less than 20, and a
+//! second coordinate larger than 700.  It is slightly faster to swap the coordinate system to
+//! treat the first coordinate as the number of rows (height), and the second as the number of
+//! columns (width), so that points adjacent by row are closer together in memory during
+//! row-major iterations.  The part two solution will need to explore beyond the bounds summed
+//! during part one; however, it is still more efficient to pre-generate the grid during the
+//! parse for sha
```
ebblake: Not sure why I left an incomplete sentence here. Will fix on the next revision.

src/year2018/day22.rs
```diff
@@ -34,58 +42,98 @@ use std::array::from_fn;
...
-type Input = [i32; 3];
-
-#[derive(Clone, Copy)]
-struct Region {
-    erosion: i32,
-    minutes: [i32; 3],
+/// Amount of slop beyond the target to include in the grid.  Too little, and this will miss
+/// paths that can take a detour to avoid a tool swap.  Too large, and this will waste time
+/// generating Regions that increase the size of the frontier explored but do not affect
+/// the shortest path.  The value picked here appears to work for all known input files, although
```
ebblake: 140 is definitely big enough for all known input files, but 30 is not ;)  Still trying to figure out if 65 is actually needed, or if I can get by with 50 (which covers the 11 inputs I found in the megathread, but not musifter's claim of 64).  The speedup of part 2 won't be as dramatic with 65 as it was with 30, but will still be more than the time added to part 1 by refactoring the grid generation into parse.

src/year2018/day22.rs
```diff
@@ -34,58 +42,98 @@ use std::array::from_fn;
...
+/// it is too small for some random cave depths.
+const SLOP_WIDTH: i32 = 10;
+const SLOP_HEIGHT: i32 = 140;
+
+pub struct Input {
+    cave: Grid<u8>, // region types for grid, (x + SLOP) by (y + SLOP)
```
ebblake: Comment is inaccurate for commit 1 applied alone, but correct after commit 2. I'll probably just merge the two commits into one.

src/year2018/day22.rs
```diff
@@ -34,58 +42,98 @@ use std::array::from_fn;
...
+    // Input gives us X,Y, but it is more efficient to use it row,column, swapped to
+    // have row-major iteration.
+    let [depth, target_row, target_col] = input.iter_signed::<i32>().chunk::<3>().next().unwrap();
+
+    let target = Point::new(target_col, target_row);
+
+    let mut row = vec![0; (target_col + SLOP_WIDTH) as usize];
+    let mut grid = Grid::new(target_col + SLOP_WIDTH, target_row + SLOP_HEIGHT, 0_u8);
+
+    // First row erosion levels are set by scaled geologic index
```
ebblake: I noticed you recently did a style cleanup pass to use trailing "." on comments. Will clean up my patch to match convention.

src/year2018/day22.rs
ebblake: No, it can't. The erosion value of the target point is explicitly NOT the product of the two neighbors, and it DOES impact at least the depth 6969 puzzle (all points southeast of the target depend on the different erosion value of the target point).  I'm using row[x] to store the erosion values across rows (cheaper to reuse the row vector than to allocate a full grid).

src/year2018/day22.rs
ebblake: Resetting grid[target] after later points have been assigned is too late. It has to be done at the point in the initialization where it can still feed later points.

src/year2018/day22.rs
ebblake: I wasn't sure about the refactor into parse, but since you like it, I'll keep it.  Parse+Part 1 is slower with the shared grid: where we used to have 40ns part1 grid creation and 500ns part2 grid creation, after the series we have a common shared 120ns [or more, with larger SLOP] grid creation that works for both part 1 and 2 (because it initializes more points than the old part 1, but fewer than the old part 2).  Part 2 is additionally sped up by fewer points being queued into A* thanks to the smaller grid. But since part 2 dominates runtime, the overall puzzle should be faster, even when I amend to a slop around 50 or 60.

src/year2018/day22.rs
```diff
@@ -34,58 +42,98 @@ use std::array::from_fn;
...
+        .map(|row| row.iter().take(*width as usize + 1).map(|point| *point as i32).sum::<i32>())
+        .sum()
 }
 
 /// A* search for the shortest path to the target.
 pub fn part2(input: &Input) -> i32 {
-    // Swap width and height.
-    let [_, height, width] = *input;
+    let Input { height, width, .. } = *input;
+    let erosion = &input.cave;
```
ebblake: I wasn't certain if this is the best syntax for getting at things. In particular, I'm converting from an incoming `erosion:Grid<u8>` with erosion type into `cave:Grid<[u16; 3]>` for the three times.  I don't know if there is a more efficient representation or way to initialize things (maybe a one-liner to express what I did in two?).  But it falls out from the fact that once I moved the generation into parse(), the erosion types should be immutable in part 2; coupled with the observation that the A* code was never consulting the erosion type because of the trick with the 0 time on the odd-on-out time.

src/year2018/day22.rs
ebblake: Most of the megathread inputs did not need ANY width slop; only depth 6969 actually went one beyond.  Keeping separate SLOP_HEIGHT=50 (or whatever we settle on?) and SLOP_WIDTH=5 will give a slightly smaller array (slightly faster parse, no real impact to A* search).

ebblake:
The following hack let me test multiple depths (takes several minutes to fail, but the failure then lists the maximum width of lots of depths):
```
diff --git i/src/year2018/day22.rs w/src/year2018/day22.rs
index 583bcd6..afdca26 100644
--- i/src/year2018/day22.rs
+++ w/src/year2018/day22.rs
@@ -58,6 +58,7 @@ pub struct Input {
     cave: Grid<u8>, // region types for grid, (x + SLOP_HEIGHT) by (y + SLOP_WIDTH)
     height: i32,    // x coordinate of the target, < 20
     width: i32,     // y coordinate of the target, > 700
+    depth: i32,
 }

@@ -94,12 +95,12 @@ pub fn parse(input: &str) -> Input {
         }
     }

-    Input { cave: grid, height: target_row, width: target_col }
+    Input { cave: grid, height: target_row, width: target_col, depth }
 }

 /// Calculate the risk level of the relevant subset of the overall cave grid.
 pub fn part1(input: &Input) -> i32 {
-    let Input { cave, height, width } = input;
+    let Input { cave, height, width, .. } = input;
     cave.bytes
         .chunks(cave.width as usize)
         .take(*height as usize + 1)
@@ -109,7 +110,7 @@ pub fn part1(input: &Input) -> i32 {

 /// A* search for the shortest path to the target.
 pub fn part2(input: &Input) -> i32 {
-    let Input { height, width, .. } = *input;
+    let Input { height, width, depth, .. } = *input;
     let erosion = &input.cave;
     let target = Point::new(width, height);

@@ -126,6 +127,7 @@ pub fn part2(input: &Input) -> i32 {
     for (i, level) in erosion.bytes.iter().enumerate() {
         cave.bytes[i][*level as usize] = 0;
     }
+    let mut parent = Grid::new(erosion.width, erosion.height, [(ORIGIN, 0); 3]);

     // Start at origin with the torch equipped.
     todo[0].push((ORIGIN, TORCH));
@@ -138,6 +140,13 @@ pub fn part2(input: &Input) -> i32 {

             // Check for completion.
             if point == target && tool == TORCH {
+                let mut max_width = 0;
+                let mut path = (target, TORCH);
+                while parent[path.0][path.1].0 != ORIGIN {
+                    max_width = max_width.max(path.0.y);
+                    path = parent[path.0][path.1];
+                }
+                println!("depth {depth}, maximum width on path: {max_width}");
                 return time;
             }

@@ -149,6 +158,7 @@ pub fn part2(input: &Input) -> i32 {
                     let heuristic = next.manhattan(target);
                     let index = (time + 1 + heuristic) as usize;

+                    parent[next][tool] = (point, tool);
                     cave[next][tool] = time + 1;
                     todo[index % BUCKETS].push((next, tool));
                 }
@@ -160,6 +170,7 @@ pub fn part2(input: &Input) -> i32 {
                     let heuristic = point.manhattan(target);
                     let index = (time + 7 + heuristic) as usize;

+                    parent[point][other] = (point, tool);
                     cave[point][other] = time + 7;
                     todo[index % BUCKETS].push((point, other));
                 }
diff --git i/tests/year2018/day22.rs w/tests/year2018/day22.rs
index 3ce095f..9697626 100644
--- i/tests/year2018/day22.rs
+++ w/tests/year2018/day22.rs
@@ -14,4 +14,13 @@ fn part1_test() {
 fn part2_test() {
     let input = parse(EXAMPLE);
     assert_eq!(part2(&input), 45);
+
+    let samples: Vec<String> = (510..12001+1).step_by(3).map(|d| format!("\
+depth: {}
+target: 15,797", d)).collect();
+    for str in samples {
+        let input2 = parse(&str);
+        part2(&input2);
+    }
+    assert_eq!(0,1);
 }
```
In the output, I see (among others) (nothing will be higher than 79 in this run, since I hard-coded target to 15,797 and had slop at 65, so the grid is only 80 columns wide):
```
depth 528, maximum width on path: 72
...
depth 594, maximum width on path: 79
depth 4404, maximum width on path: 79
depth 6909, maximum width on path: 78
```

src/year2018/day22.rs
maneatingape: `let &Input { height, width, cave: ref erosion } = input;` is cleaner and will dereference the `i32`s while forcing cave to be bound to a reference.

This will trigger the `ref-patterns` lint, please change from `warn` to `allow`.

src/year2018/day22.rs
ebblake: Testing shows that the width slop can stay 5 or less, while the height slop needs to be at least 64 for musifter's input value (I won't repeat it here since it is not public, but I did enhance the testsuite with a similar depth value).

ebblake:
I've now confirmed with musifter offline that the choice of SLOP_HEIGHT = 65 works for his input, while dropping to 60 would fail. To respect the fact that musifter's input does not appear to be leaked in the megathread, I'm not leaking it here either, but I did add a (different) depth value to the testsuite that has similar properties.

PR #65, opened 2026-04-23
Exploit niceness of input file for faster search

ebblake:
## Description

As observed on the megathread, the input files are nice, in that the theoretical maximum overlap when flattening the Manhattan intersection ranges of each nanobot in isolation down to one dimension happens to also coincide with the actual maximum overlap.  While a generic solution still has to rely on a binary search of refining a search region with an O(n) check for number of intersections at each candidate point being searched, we can get by with just one sorted pass over the nanobots.

On my machine, execution speeds up from 1.5ms to 40us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

  specific intrinsics.
- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2018/day23.rs
+++ b/src/year2018/day23.rs
@@ -1,8 +1,8 @@
 //! # Experimental Emergency Teleportation
 //!
-//! Part two implements a 3D version of binary search. Starting with a single cube that encloses all
-//! nanobots, each cube is further split into 8 smaller cubes until we find the answer.
-//! Cubes are stored in a [`MinHeap`] ordered by:
+//! A generic solution to part two would be a 3D version of binary search. Starting with a single
+//! cube that encloses all nanobots, each cube is further split into 8 smaller cubes until we find
+//! the answer. Cubes can be stored in a [`MinHeap`] ordered by:
 //!
 //! * Greatest number of nanobots in range.
 //! * Least distance to origin.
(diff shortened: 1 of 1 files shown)


src/year2018/day23.rs
```diff
@@ -15,8 +15,16 @@
 //! * There are no cubes that are closer.
 //! * The coordinates cannot be refined any further.
 //!
-//! [`MinHeap`]: crate::util::heap
-use crate::util::heap::*;
+//! However, the actual input files lend themselves to an even faster answer. At the high level,
```
maneatingape: Nits:
* "At the high level" =>  "At a high level"
* "a given nanobot  " => One trailing space
* "best point may have less" => "best point may have fewer"

src/year2018/day23.rs
```diff
@@ -110,23 +55,30 @@ pub fn part1(input: &[Nanobot]) -> usize {
...
-        for next in cube.split() {
-            let in_range = input.iter().filter(|nb| next.in_range(nb)).count();
-            let key = (input.len() - in_range, next.closest(), next.size());
-            heap.push(key, next);
+    // Start by populating the possible distances that can reach each nanobot.
+    let mut endpoints = Vec::with_capacity(2_000);
+    let origin = Nanobot::from([0, 0, 0, 0]);
+    for bot in input {
+        let dist = bot.manhattan(&origin);
+        let low = if dist > bot.r { dist - bot.r } else { 0 };
```
maneatingape: Simplifies to `let low = (dist - bot.r).max(0);`

src/year2018/day23.rs
```diff
@@ -110,23 +55,30 @@ pub fn part1(input: &[Nanobot]) -> usize {
...
-        }
-
-        for next in cube.split() {
-            let in_range = input.iter().filter(|nb| next.in_range(nb)).count();
-            let key = (input.len() - in_range, next.closest(), next.size());
-            heap.push(key, next);
+    // Start by populating the possible distances that can reach each nanobot.
+    let mut endpoints = Vec::with_capacity(2_000);
+    let origin = Nanobot::from([0, 0, 0, 0]);
+    for bot in input {
```
maneatingape, line 62: Nit: Blank line between declarations and loop

src/year2018/day23.rs
```diff
@@ -110,23 +55,30 @@ pub fn part1(input: &[Nanobot]) -> usize {
...
+        endpoints.push((low, 1));
+        endpoints.push((dist + bot.r + 1, -1));
+    }
+    endpoints.sort_unstable();
+
+    // Determine the distance that has the maximum overlap in ranges.
+    let mut best_dist = 0;
+    let mut best_total = 0;
+    let mut total = 0;
+    for (dist, delta) in endpoints {
```
maneatingape, line 75: Nit: Blank line between declarations and loop.

maneatingape (changes requested):
Looks good, much faster and the code is simpler.

maneatingape (approved):

PR #66, opened 2026-05-06
Improve caching of hits

ebblake:
## Description

The grid size for seen was twice as large as needed: since we pair points in reading order, delta.y will always be positive, so we don't need 2*height in the seen grid.  On the other hand, computing the gcd is only part of the effort needed to get from a particular delta back to the adjusted index that serves as the witness for the first visible asteroid along an integer-coordinate line.  Reworking the code to move more computation into the cache population, so that the cache lookup bypasses more work, provides a nice speed boost.  On my input, I count over 53 thousand point deltas, but only populate around 800 entries in the cache.  The updated cache needs double-width (the same as the seen grid), so I'm not really reducing memory usage, just using it more effectively.

On my machine, this improves runtime from around 660-790us to a faster 365-440us (the runtimes are quite variable; probably because there is enough memory in play to not comfortably fit in L1 cache).

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day10.rs
+++ b/src/year2019/day10.rs
@@ -71,8 +71,8 @@ pub fn parse(input: &str) -> Input {
     }
 
     // Find asteroid with the highest visibility.
-    let mut seen = Grid::new(2 * width, 2 * height, 0);
-    let mut cache = grid.same_size_with(0);
+    let mut seen = Grid::new(2 * width, height, 0);
+    let mut cache = seen.same_size_with(ORIGIN);
     let mut visible = vec![0; points.len()];
     let mut max_visible = 0;
     let mut max_index = 0;
@@ -80,19 +80,19 @@ pub fn parse(input: &str) -> Input {
     for i in 0..(points.len() - 1) {
         for j in (i + 1)..points.len() {
             let delta = points[j] - points[i];
-            let key = Point::new(delta.x.abs(), delta.y.abs());
+            // Because points was populated in order from left to right and top to bottom, delta.y will
+            // be non-negative; delta.x might be negative, but shifting it by width gives us a positive
+            // reference point that can then cache the mapping to a representative index.
+            let key = Point::new(width + delta.x, delta.y);
 
-            if cache[key] == 0 {
-                cache[key] = key.x.gcd(key.y);
+            if cache[key] == ORIGIN {
+                // Key insight is that points on the same line are integer multiples of each other.
+                let gcd = delta.x.abs().gcd(delta.y);
+                let adjusted = Point::new(delta.x / gcd + width, delta.y / gcd);
+
+                cache[key] = adjusted;
             }
-            let gcd = cache[key];
-
-            // Key insight is that points on the same line are integer multiples of each other.
-            let adjusted = Point::new(delta.x / gcd, delta.y / gcd);
-
-            // This works as the points are in order from left to right and top to bottom,
-            // so we process points from nearest to furthest.
-            let index = Point::new(width + adjusted.x, height + adjusted.y);
+            let index = cache[key];
 
             if seen[index] <= i {
                 seen[index] = i + 1;


ebblake:
Possible idea for memory savings: instead of caching Points (a struct with two i32, and therefore 8 bytes), we can exploit that the grid is decently small (mine was 21x21, although I know at least one other person got a 36x36) that converting grid.index(point) to u16 for storage, and then u16 back to usize for cache[index] lookups, should work.

ebblake:
> Possible idea for memory savings: instead of caching Points (a struct with two i32, and therefore 8 bytes), we can exploit that the grid is decently small (mine was 21x21, although I know at least one other person got a 36x36) that converting grid.index(point) to u16 for storage, and then u16 back to usize for cache[index] lookups, should work.

No real difference with this patch:
```
diff --git i/src/year2019/day10.rs w/src/year2019/day10.rs
index bbaf8d4..4236266 100644
--- i/src/year2019/day10.rs
+++ w/src/year2019/day10.rs
@@ -50,9 +50,27 @@ use crate::util::grid::*;
 use crate::util::math::*;
 use crate::util::point::*;
 use std::cmp::Ordering;
+use std::ops::Index;
+use std::ops::IndexMut;

 type Input = (i32, i32);

+impl<T> Index<u16> for Grid<T> {
+    type Output = T;
+
+    #[inline]
+    fn index(&self, index: u16) -> &Self::Output {
+        &self.bytes[index as usize]
+    }
+}
+
+impl<T> IndexMut<u16> for Grid<T> {
+    #[inline]
+    fn index_mut(&mut self, index: u16) -> &mut Self::Output {
+        &mut self.bytes[index as usize]
+    }
+}
+
     let grid = Grid::parse(input);
     let width = grid.width;
@@ -72,7 +90,7 @@ pub fn parse(input: &str) -> Input {

     // Find asteroid with the highest visibility.
     let mut seen = Grid::new(2 * width, height, 0);
-    let mut cache = seen.same_size_with(ORIGIN);
+    let mut cache = seen.same_size_with(0_u16);
     let mut visible = vec![0; points.len()];
     let mut max_visible = 0;
     let mut max_index = 0;
@@ -85,12 +103,12 @@ pub fn parse(input: &str) -> Input {
             // reference point that can then cache the mapping to a representative index.
             let key = Point::new(width + delta.x, delta.y);

-            if cache[key] == ORIGIN {
+            if cache[key] == 0 {
                 // Key insight is that points on the same line are integer multiples of each other.
                 let gcd = delta.x.abs().gcd(delta.y);
-                let adjusted = Point::new(delta.x / gcd + width, delta.y / gcd);
+                let adjusted = Point::new(width + delta.x / gcd, delta.y / gcd);

-                cache[key] = adjusted;
+                cache[key] = (adjusted.y * cache.width + adjusted.x) as u16;
             }
             let index = cache[key];

```

ebblake:
Another possible speed up, this time for part 2. As written, if there are 350 points, part 2 sorts ALL of them, O(n log n) for n = 350.  But since you are already doing an O(n) pass to normalize the points in relation to the station, it is possible to do two half-grid traversals (one from station-1 down to 0, the other from station+1 up to max index) to determine which points are in range on the first 360 swing around the field, while simultaneously sorting them into four buckets based on quadrant.   If the part 1 answer is higher than 200 (which mine was), then that plus the count of items in each of the 4 buckets is all the more you need to know WHICH bucket needs angle sorting; if not, then you do a second (or more) O(n) sweeps until finding the right round.

If we were worried about part 1 regularly being under 200 for some inputs, your code for simultaneously determining the bucket number of a point based on its distance-squared metric during a single O(n) pass is viable - but if most valid inputs have a part 1 larger than 200, then you don't even need to compute distance, because sorting by angles once you have the right bucket is adequate.  But the idea of filtering into buckets in O(n) before doing an angle sort means that your angle sort is then O(n log n) for an n closer to 50, rather than 350, making it faster.

maneatingape:
This is great, happy to merge as-is, unless there was another idea you wanted to explore first?

ebblake:
> then you don't even need to compute distance, because sorting by angles once you have the right bucket is adequate. But the idea of filtering into buckets in O(n) before doing an angle sort means that your angle sort is then O(n log n) for an n closer to 50, rather than 350, making it faster.

In fact, since we know we are looking for the nth item in the bucket, we can use quickselect, to try and get its typical O(n) behavior (worst case O(n^2) - but the gist of quickselect is that it partitions the list into hopefully two equal sets (like quicksort), but then only traverses into the half of the list that contains n)


> This is great, happy to merge as-is, unless there was another idea you wanted to explore first?

I can do explorations on quickselect as a separate patch.

ebblake:
I implemented quickselect for part 2 in my m4 solution, and it DID make a difference there. However, as a very quick test, I separated part2() out of parse() to see where the time is being spent:
```
diff --git i/src/year2019/day10.rs w/src/year2019/day10.rs
index bbaf8d4..71abcdf 100644
--- i/src/year2019/day10.rs
+++ w/src/year2019/day10.rs
@@ -51,7 +51,11 @@ use crate::util::math::*;
 use crate::util::point::*;
 use std::cmp::Ordering;

-type Input = (i32, i32);
+pub struct Input {
+    part_one: i32,
+    max_index: usize,
+    points: Vec<Point>,
+}

     let grid = Grid::parse(input);
@@ -107,9 +111,19 @@ pub fn parse(input: &str) -> Input {
         }
     }

+    Input { part_one: max_visible, max_index, points }
+}
+
+    input.part_one
+}
+
+pub fn part2(input: &Input) -> i32 {
+    let mut points = input.points.clone();
+
     // Remove our new base of operations, then sort remaining asteroids in clockwise order to
     // group by angle.
-    let station = points.swap_remove(max_index);
+    let station = points.swap_remove(input.max_index);
     points.iter_mut().for_each(|p| *p -= station);
     points.sort_unstable_by(|&a, &b| clockwise(a, b));

@@ -134,17 +148,7 @@ pub fn parse(input: &str) -> Input {

     // The 200th asteroid is at index 199.
     let target = station + points[groups[199].2];
-    let result = 100 * target.x + target.y;
-
-    (max_visible, result)
-}
-
-pub fn part1(input: &Input) -> i32 {
-    input.0
-}
-
-pub fn part2(input: &Input) -> i32 {
-    input.1
+    100 * target.x + target.y
 }

 /// The [`then`] method chains [`Ordering`] results.
```

and `cargo bench year2019::day10` reports:
```
test year2019::day10::parse_bench ... bench:     429,101.07 ns/iter (+/- 11,333.15)
test year2019::day10::part1_bench ... bench:           1.40 ns/iter (+/- 0.05)
test year2019::day10::part2_bench ... bench:      12,790.09 ns/iter (+/- 3,055.16)
```

So even if quickselect does help part 2, the effect on the overall day will be pretty minimal.

PR #67, opened 2026-05-07
Use a faster quickselect rather than full sort

ebblake:
## Description

Followup to #66.

Sorting 350+ elements to finding the 200th in sorted order is minimum O(n log n); plus, our sort criteria was rather complex.  It is possible to simplify things by working a quadrant at a time (reusing the cache from part 1) and then use quickselect (often as fast as O(n)) on a smaller list (around 50 elements), at which point we no longer need to compute Euclidean distance as a sort tie-breaker.

Runtime improvement is minimal in relation to part 1.  On my machine, a pre-patch isolation of part2 to a separate function only took 12us, and I often had more than that much variability in the timings for parse near 400us.  And since this patch reuses cache from part 1, I can't easily separate it out for post-patch timing analysis.  But I was able to add temporary debug to angle() to prove that pre-patch called angle() 3131 times, and post-patch only 298.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day10.rs
+++ b/src/year2019/day10.rs
@@ -130,10 +130,9 @@ pub fn parse(input: &str) -> Input {
         groups.push((first, second, i));
     }
 
-    groups.sort_unstable();
-
-    // The 200th asteroid is at index 199.
-    let target = station + points[groups[199].2];
+    // The 200th asteroid is at index 199. Quickselect is faster than a full sort.
+    let index = groups.select_nth_unstable(199).1.2;
+    let target = station + points[index];
     let result = 100 * target.x + target.y;
 
     (max_visible, result)


ebblake:
Full disclosure - I relied heavily on Google Gemini AI for coming up with the `Box<dyn Iterator`... syntax for how to provide an enumerated iterator that produces just the points in a quadrant while still maintaining the overall index and while mixing `Filter<Iter>` and `Filter<Rev<Iter>>` in the same match statement (ie. I knew what I wanted, but not how to express it).

maneatingape:
> Full disclosure - I relied heavily on Google Gemini AI for coming up with the `Box<dyn Iterator`... syntax for how to provide an enumerated iterator that produces just the points in a quadrant while still maintaining the overall index and while mixing `Filter<Iter>` and `Filter<Rev<Iter>>` in the same match statement (ie. I knew what I wanted, but not how to express it).

That's totally fine and and reasonable use of a tool. Claude Code is also great, IMHO the strongest model at the moment.

maneatingape:
This is slightly faster, however using quickselect with the existing code gives the same timings and is much simpler:

```
    // The 200th asteroid is at index 199.
    let index = groups.select_nth_unstable(199).1.2;
    let target = station + points[index];
```

ebblake:
Sure, the simpler patch makes sense (tracing calls to angle doesn't drop, because the time being saved is now in fewer calls to the comparator for the second sort of groups).  But I did see one other minor speedup: the tie-breaker for distance is only ever reached after quadrant and angle are equal - which means it can do less math (comparing absolute values of either x or y, rather than performing multiplication).

src/year2019/day10.rs
```diff
@@ -173,8 +169,8 @@ fn angle(point: Point, other: Point) -> Ordering {
     (other.x * point.y).cmp(&(other.y * point.x))
 }
 
-/// Euclidean distance squared. No need to take square root since we're only interested
-/// in the *relative* distance.
-fn distance(point: Point) -> i32 {
-    point.x * point.x + point.y * point.y
+/// Since this is only reached for points in the same quadrant and angle, comparing absolute value of
```
maneatingape: This will treat points with x = 0 the same, no matter what y value.
Benchmarking show no difference for me, so let's revert to the previous version.

maneatingape (changes requested):

src/year2019/day10.rs
ebblake: Good point. Manhattan distance would avoid the problem while still reducing multiplies, but since it isn't the hot point, I will roll it back when I get a chance.

maneatingape (approved):

PR #69, opened 2026-05-14
Use a dynamic programming approach

ebblake:
## Description

Rather than run 1200 Intcode machines (5! permutations * 5 machines * 2 parts) and 240 comparisons (5! * 2 parts, O(n!) work), this problem can be solved with 15 Intcode machines (2 inputs * 5 machines for the 5 Ax+B constants in part 1, 1 input * 5 machines for the 5\*10 +1/+2/\*2 constants in part 2), followed by 160 comparisons (average 5/2 bits per set * 2^5 sets * 2 parts, O(n*2^n) work).

Massively speeds up the runtime.  On my machine, it drops from 30 and 150us for parts 1 and 2, to instead being 1.0 and 2.5us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day07.rs
+++ b/src/year2019/day07.rs
@@ -1,68 +1,127 @@
 //! # Amplification Circuit
 //!
-//! Brute force solution for both parts using the utility [`permutations`] method to test each of
-//! the possible 5! or 120 permutations of the phase settings.
+//! The biggest savings to be had here are by realizing that each machine id runs an independent
+//! linear equation on the input. For part 1, these equations are in the form Ax + B, where B can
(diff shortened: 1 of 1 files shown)


maneatingape:
Intcode problems are interesting becuase the question is where do you draw the line on analysis of the program. At one end you have treating the program as a black box, the other end is peephole analysis of magic constants.

My preference for 2019 is to treat the program as a black box. This is not always the fastest, but feels the most in the spirit of the challenge. Other folks can make the perfectly valid choice to analyse the program more.

This is too far from a black box to fit in with the meta of these solutions.

PR #70, opened 2026-05-15
Use Held-Karp dynamic programming rather than permutations

ebblake:
## Description

Experiment with a different algorithm for finding the shortest path, using O(n^2*2^n) rather than O(n!) work.  For a given part, the old code had 2520 callbacks from half_permutations with 8 calls to .max per callback; the new code has only 3584 total calls to .max, and with less overhead than what permutations required for shuffling data around.

I would have loved to have iterated on 3..127 instead of 3..255; that produces only 1344 calls to .max, but only works on inputs where dropping the shortest edge from the longest cycle still produces the longest path.

On my laptop, performance improves from about 40us to 12us. I also tried separating the two tables (doing Held-Karp in part1() and part2() rather than joint iteration in parse(), but that had more overhead with duplicated set traversal.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2015/day09.rs
+++ b/src/year2015/day09.rs
@@ -1,30 +1,55 @@
 //! # All in a Single Night
 //!
-//! This is a variant of the classic NP-hard
-//! [Travelling Salesman Problem](https://en.wikipedia.org/wiki/Travelling_salesman_problem).
+//! This is a variant of the classic NP-hard [Travelling Salesman Problem].
 //!
 //! There are 8 locations, so naively it would require checking 8! = 40,320 permutations. We can
 //! reduce this to 7!/2 = 2,520 permutations by arbitrarily choosing one of the locations as the
 //! start, and skipping lexically reversed permutations (since the path a->b->c has the same
-//! length as c->b->a).
+//! length as c->b->a). Computing the shortest and longest path is then done by completing the
+//! cycle for each permutation, then discarding the longest or shortest edge seen along the way.
+//! Skipping lexically reversed permutations is possible with
+//! [Steinhaus-Johnson-Trotter][Steinhaus-Johnson-Trotter's algorithm].
 //!
-//! We then compute the distance to complete the trip and return to the original location.
-//! Since the problem does not ask us to end up in the same location we then "break" the cycle.
-//! To compute the shortest journey we remove the longest single journey and to compute the
-//! longest journey we remove the shortest single journey.
+//! However, since the graph is complete (every node has a distance to every other node), this
+//! particular problem can be solved even faster, by avoiding the overhead of permutations and
+//! instead using [Held-Karp's dynamic programming][Held-Karp] solution. This algorithm is
+//! O(n²*2ⁿ); for our puzzle with 8 nodes, this gives `8*7/2*256/2` or 3,584 comparisons needed.
+//! On the surface, this is more comparisons than the 2,520 sequences visited by the O(n!)
+//! permutation solution, but set manipulation is less expensive than computation of permutations
+//! followed by casting out the longest or shortest edge, so it is an overall win.
+//!
+//! The core behavior of Held-Karp involves computing the function g(set, k) for all possible
+//! sets of cities, where the function represents the best (shortest or longest) distance seen so
+//! far for a given set of cities and ending on the city k which is a member of the set. Unlike the
+//! permutations approach which visits every path in the graph, Held-Karp discards information
+//! along the way to compute only the best cycle in a graph anchored to a given start point. It is
+//! easy to demonstrate a graph where the best path is not part of the best cycle (discarding the
+//! longest edge from the shortest cycle might leave you with a path longer than the graph's true
+//! shortest path, if that other path had a different start anchor). But this is easy to work
+//! around, by adding a ninth "location" with distance 0 to every other location, and using that
+//! location as the start and end for every cycle, at which point the best cycle of nine includes
+//! the best path of all eight locations.
+//!
+//! The algorithm is recursive: with a base case of g({k},k) being zero (the best distance
+//! to a singleton set from our ninth point is 0), all other g(set,k) can be computed by
+//! iterating over each member of set excluding k, and choosing the best variant g(set∖k,m)+d(k,m)
+//! possible from a smaller set ending in m. Iterating over bitmasks from 0 to 255 ensures
+//! that all earlier subsets are available when computing for a larger set.
 //!
 //! For speed we first convert each location into an index, then store the distances between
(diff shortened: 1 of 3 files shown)


ebblake:
If you like this, 2015 day 13 and 2016 day 24 could use the same treatment.  At which point, there would no longer be any clients of half_permutations in crate::util::slice; see also #69 that gets rid of permutations from 2019 day 7.

src/year2015/day09.rs
```diff
@@ -1,30 +1,52 @@
...
-//! [Travelling Salesman Problem](https://en.wikipedia.org/wiki/Travelling_salesman_problem).
+//! This is a variant of the classic NP-hard [Travelling Salesman Problem].
 //!
 //! There are 8 locations, so naively it would require checking 8! = 40,320 permutations. We can
 //! reduce this to 7!/2 = 2,520 permutations by arbitrarily choosing one of the locations as the
 //! start, and skipping lexically reversed permutations (since the path a->b->c has the same
-//! length as c->b->a).
+//! length as c->b->a). Computing the shortest and longest path is then done by completing the
+//! cycle for each permutation, then discarding the longest or shortest edge seen along the way.
+//! Skipping lexically reversed permutations is possible with [Steinhaus-Johnson-Trotter's algorithm].
```
ebblake: Not sure I got the markdown correct for forward-references of a hyperlink in the body vs. the URL later in the document, compared to the older inline URL.

src/year2015/day09.rs
```diff
@@ -1,30 +1,52 @@
...
 //!
-//! We then compute the distance to complete the trip and return to the original location.
-//! Since the problem does not ask us to end up in the same location we then "break" the cycle.
-//! To compute the shortest journey we remove the longest single journey and to compute the
-//! longest journey we remove the shortest single journey.
+//! However, since the graph is complete (every node has a distance to every other node), this
+//! particular problem can be solved even faster, by avoiding the overhead of permutations and
+//! instead using [Held-Karp's dynamic programming][Held-Karp] solution. This algorithm is
+//! O(n²*2ⁿ); for our puzzle with 8 nodes, this gives `8*7/2*256/2` or 3,584 comparisons needed.
+//! On the surface, this is more comparisons than the O(n!) permutation solution 2,520 comparisons,
```
ebblake: possible wording touchup:
more comparisons than the O(n!) 2,520 comparisons of the permutation solution

src/year2015/day09.rs
```diff
@@ -50,42 +72,45 @@ pub fn parse(input: &str) -> Result {
         distances[stride * end + start] = distance;
     }
 
-    let mut global_min = u32::MAX;
-    let mut global_max = u32::MIN;
-    let mut indices: Vec<_> = (1..stride).collect();
+    // Initialize a table for each part: 2ⁿ sets with n distances per set. Default 0 matches
+    // g({k},k) of zero for all singleton sets.
+    let mut table_one = vec![vec![0_u16; stride]; 1 << stride];
+    let mut table_two = vec![vec![0_u16; stride]; 1 << stride];
```
ebblake: table_{one,two} should be one-dimensional, the same as distances. Doing that makes another nice speedup.

src/year2016/day24.rs
```diff
@@ -77,26 +76,50 @@ pub fn parse(input: &str) -> Input {
     }
 
     // Solve both parts simultaneously.
-    let mut part_one = u32::MAX;
-    let mut part_two = u32::MAX;
-    let mut indices: Vec<_> = (1..found.len()).collect();
+    // Initialize a table for each part: 2ⁿ⁻¹ sets with n-1 distances per set. Default each g({k},k)
+    // singleton to distance[0][k+1] (since bit 0 maps to node 1), while the initial value of other
+    // sets does not matter.
+    let mut table = [[0_u16; 7]; 1 << 7];
```
ebblake, line 82: Day 9 and 13 used vec for the distance grid because the tests had fewer nodes, so I did likewise for the table. Day 24 uses an array for the distance grid, the test failure comes from having fewer nodes without being prepared for it.

ebblake: I like the 2D array a bit better (it's nice for table[set][m]); oversized the table to 2^11 entries when the test only needs 5*2^5 entries is not too bad, so the fix may just be clamping the iterations to the actual node count even when the table is oversized, rather than converting to 1D vec

ebblake: Test failure fixed in the latest iteration, but I'm still torn between whether the 2D array (exact size for puzzle but oversized for test cases) vs. 1D vector (always exact size) is better, and whether the three puzzles should be consistent in using the same style.

ebblake:
Reddit writeup on the speedups gained by this patch series: https://www.reddit.com/r/adventofcode/comments/1te7ayd/2015_day_9rust_solving_tsp_faster_than/

src/year2015/day09.rs
```diff
@@ -49,42 +75,44 @@ pub fn parse(input: &str) -> Result {
...
+    let mut table_one = vec![0_u16; stride * (1 << stride)];
+    let mut table_two = vec![0_u16; stride * (1 << stride)];
 
-    indices.half_permutations(|slice| {
-        let mut sum = 0;
-        let mut local_min = u32::MAX;
-        let mut local_max = u32::MIN;
+    // Visit each non-empty set in order, with no work to do for singleton sets.
+    for set in 3_usize..(1 << stride) {
+        if set & !(set - 1) == set {
```
maneatingape: To make the intent clearer bit twiddling hack can be replaced with either `if set.is_power_of_two()` or `if set.count_ones() < 2`

(Same for other 2 call sites)

src/year2015/day09.rs
```diff
@@ -49,42 +75,44 @@ pub fn parse(input: &str) -> Result {
...
+    // g({k},k) of zero for all singleton sets. Initial value of other sets does not matter.
+    let mut table_one = vec![0_u16; stride * (1 << stride)];
+    let mut table_two = vec![0_u16; stride * (1 << stride)];
 
-    indices.half_permutations(|slice| {
-        let mut sum = 0;
-        let mut local_min = u32::MAX;
-        let mut local_max = u32::MIN;
+    // Visit each non-empty set in order, with no work to do for singleton sets.
+    for set in 3_usize..(1 << stride) {
```
maneatingape, line 84: Minor nit: Add a small extra comment here, e.g. "Starting from 3 since 1 and 2 are singleton sets`

(Same for other 2 call sites)

maneatingape (changes requested):
Another nice patch with a great speedup and the code is equally simple to understand.

maneatingape (approved):
👍

PR #72, opened 2026-05-16
Use shorter springdroid script

ebblake:
## Description

Every line of springdroid script causes that much more execution time for the intcode program (looking at Eric's sample source code [1], it blindly executes a sequence of intcode instructions for every compiled line of springscript on every bit of each 9-bit hole pattern, even if extra script instructions don't change the final contents of register J). In tracing things, I see about 50k fewer Intcode opcodes executed for each line of springdroid eliminated.  Thus, coming up with a shorter program that still passes all inputs is worthwhile.

[1] https://github.com/topaz/aoc2019-intcode/blob/master/samples/day21.intcode

On my laptop, runtime improved from 145us/3.4ms for part1/part2 pre-patch, to 125us/2.8ms post-patch.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day21.rs
+++ b/src/year2019/day21.rs
@@ -1,7 +1,9 @@
 //! # Springdroid Adventure
 //!
 //! Jumps are always 4 tiles wide, landing on `D`. If needed we can jump again immediately
-//! landing on `H`.
+//! landing on `H`. The intcode program runs faster if the springscript is shorter, so although
+//! multiple scripts work, any solution that treats intcode as a black box will be better if
+//! the script is minimized.
 //!
 //! ## Part One
 //!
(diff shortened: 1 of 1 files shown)


ebblake:
My ultimate solution in m4 for day 21 was to break the black box, where I decompiled enough of the code to figure out where in memory the 9-bit hole patterns are stored, as well as how the score increments for each hole pattern if the script were to be successful. My solution then computes the same score without ever executing a single instruction of intcode. That approach would DRASTICALLY speed up this day (and be in line with some of your other years' VM solutions by disassembly doing the same math in native Rust rather than emulation) - but it would be the first such Intcode program where your solution would break the black box. So I'm not sure if you want to go with that approach. Of course, if you DO break the black box, there are several other intcode days where I can also solve quite a bit faster by sniffing magic numbers out of the input file rather than actually executing intcode.

ebblake:
If it helps, in my analysis of all input files that I have seen so far (including Eric's sample program), the hole patterns are represented as 8-bit values, where 1 is ground and 0 is a hole, but the code concatenates an implicit 9th bit hole in front. Part 1 appears to universally be hole pattern 255, 63, then a random permutation of the five patterns 95, 127, 191, 159, and 223, which makes sense: all of them have the last 5 bits as ground (since there aren't enough registers to see that far when WALKing), and this leaves you with the remaining patterns 0b0111, 0b0001, 0b0010, 0b0011, 0b0100, 0b0101, 0b0110. As this is all possible 4-bit patterns with a leading 0 but not 4 consecutive 0, it is easy enough to prove that my 4-instruction springscript can pass all 7 of those patterns even without using B (there are a few other 4-line programs that pass, but yours is easy to explain). Part 2 has a random permutation of 153 patterns that must succeed, but again I have seen reddit posts that exhaustively prove that all such patterns can pass with a 6-line script (several variations of a passing script exist, although I'm partial to the version that didn't need to use T), but no 5-line script can pass all of them.

maneatingape (approved):
Neat!

PR #73, opened 2026-05-18
Use better slope estimate and geometry for less intcode

ebblake:
## Description

Cut the number of test() runs of intcode (on my input) from 453 to
just 150, by spending a little bit of time during parse to refine the
slope estimate for fewer false positives in part 1, and then by using
that more accurate slope to start part 2 much closer to the actual
answer.
    
On my laptop, this cuts parse/part1/part2 runtime from 200/215/540us
to 97/199/42us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day19.rs
+++ b/src/year2019/day19.rs
@@ -1,40 +1,73 @@
 //! # Tractor Beam
 //!
-//! Finds the approximate boundary of the upper and lower edges of the beam expressed as a slope.
-//! We then skip the relatively expensive intcode test if the x and y coordinates lie outside.
+//! The intcode program computes a linear inequality: returning true if an integer point lies on
+//! or between two lines through the origin, often with irrational slope. The intcode program was
+//! designed so that the two lines are close enough that there are no integer solutions when `y=1`,
+//! so there are intentionally one or two discontinuities between the origin and the bulk of the
+//! beam. This solution finds the approximate boundary of the upper and lower edges of the beam
+//! expressed as an integer ratio for slope. We then skip the relatively expensive intcode test if
+//! the x and y coordinates lie outside. Once we identify an edge past the initial discontinuities,
+//! scaling along the lines buys more accuracy and thus fewer later intcode runs.
+//!
+//! For part 2, we can further speed up the process by using geometry to hone in on a viable
+//! target to start searching at. Our target point `(x,y)` is related to our two slopes as:
+//! ```none
+//!   scale*y = upper*(x+99)
+//!   scale*x = lower*(y+99)
+//! ```
+//! Those two equations can be represented in matrix form:
+//! ```none
+//!   [upper-scale][x] = [-99*upper]
+//!   [scale-lower][y] = [ 99*lower]
+//! ```
+//! where inverting the matrix gives a solution:
+//! ```none
+//!   determinant = scale * scale - lower * upper
+//!   x = 99 * (lower * upper + lower * scale) / determinant
+//!   y = 99 * (lower * upper + upper * scale) / determinant
+//! ```
 use super::intcode::*;
 use crate::util::parse::*;
 
 pub struct Input {
     code: Vec<i64>,
-    lower: i64,
-    upper: i64,
+    scale: i64,
+    lower: i64, // The slope scale/lower just outside left boundary.
+    upper: i64, // The slope upper/scale just outside right boundary.
 }
 
 pub fn parse(input: &str) -> Input {
+    // Pick an initial scale large enough to be past the discontinuities for all known inputs.
     let code: Vec<_> = input.iter_signed().collect();
-    let mut lower = 0;
-    let mut upper = 0;
+    let mut lower = 1;
+    let mut upper = 1;
+    let mut scale = 5;
 
-    // Find slope of lower and upper edges, rounding down to prevent false negatives.
-    while !test(&code, lower + 1, 50) {
-        lower += 1;
-    }
-    while !test(&code, 50, upper + 1) {
-        upper += 1;
+    // Find approximate slope of lower and upper edges, rounding down to prevent false negatives.
+    // Each scaling iteration adds another bit of accuracy to our approximation.
+    while scale < 1024 {
+        scale *= 2;
+        lower *= 2;
+        upper *= 2;
+        while !test(&code, lower + 1, scale) {
+            lower += 1;
+        }
+        while !test(&code, scale, upper + 1) {
+            upper += 1;
+        }
     }
 
-    Input { code, lower, upper }
+    Input { code, scale, lower, upper }
 }
 
 pub fn part1(input: &Input) -> i64 {
-    // Handle origin specially as the slope check excludes it.
-    let mut result = test(&input.code, 0, 0) as i64;
+    // The origin is always set, but no other point occurs on that row or column.
+    let mut result = 1;
 
-    // The beam is continuous so we only need to find the left and right edges.
-    for y in 0..50 {
-        let left = (0..50).find(|&x| inside(input, x, y));
-        let right = (0..50).rfind(|&x| inside(input, x, y));
+    // Scan all remaining points; this works even on lines with no integer hits.
+    for y in 1..50 {
+        let left = (1..50).find(|&x| inside(input, x, y));
+        let right = (left.unwrap_or(50)..50).rfind(|&x| inside(input, x, y));
         if let Some((left, right)) = left.zip(right) {
             result += right - left + 1;
         }
@@ -44,8 +77,10 @@ pub fn part1(input: &Input) -> i64 {
 }
 
 pub fn part2(input: &Input) -> i64 {
-    let mut x = 0;
-    let mut y = 0;
+    // See comments above about derivation of initial guess for x and y.
+    let determinant = input.scale * input.scale - input.lower * input.upper;
+    let mut x = 99 * (input.lower * input.upper + input.lower * input.scale) / determinant;
+    let mut y = 99 * (input.lower * input.upper + input.upper * input.scale) / determinant;
     let mut moved = true;
 
     // Increase the right and bottom edges of our box until they are both inside the beam.
@@ -69,7 +104,9 @@ pub fn part2(input: &Input) -> i64 {
 /// Skip the relatively expensive intcode test if the point lies outside the beam's slopes.
 /// The slope check has some false positives but no false negatives.
 fn inside(input: &Input, x: i64, y: i64) -> bool {
-    50 * y > input.upper * x && 50 * x > input.lower * y && test(&input.code, x, y)
+    input.scale * y > input.upper * x
+        && input.scale * x > input.lower * y
+        && test(&input.code, x, y)
 }
 
 /// Definitive but slower check.


src/year2019/day19.rs
```diff
@@ -2,26 +2,65 @@
...
+//! Once we identify an edge, scaling from 50 to 1600 buys more accuracy and thus fewer later
+//! intcode runs.
+//!
+//! For part 2, we can further speed up the process by using geometry to hone in on a viable
+//! target to start searching at. If we consider a scaled-down beam with target point A on slope
+//! upper as (x1, y1), and point B on slope lower as (x2, y2), the bounding box has a top left
+//! corner at point S at (x2, y1):
+//! ```
+//!  1
+//!  11
```
ebblake: Based on the test failure, I'm formatting this markdown incorrectly

src/year2019/day19.rs
```diff
@@ -2,26 +2,65 @@
...
+//!         111SxxA11  <--- (x1, y1)
+//!          11xxxx1111
+//!           1xxxx111111
+//!  (x2, y2)> Bxxx11111111
+//!             1111111111111
+//! ```
+//!
+//! For a 100x100 box, we have a system of equations x1=x2+99, and y2=y1+99, which we solve as:
+//! ```
+//! x2 = ((upper/1600 * 99) + 99) / (1600/lower - upper/1600)
```
ebblake: I ought to rewrite this as m1=upper/scale rather than a hard-coded constant

src/year2019/day19.rs
```diff
@@ -2,26 +2,65 @@
...
+//! ```
 use super::intcode::*;
 use crate::util::parse::*;
 
 pub struct Input {
     code: Vec<i64>,
-    lower: i64,
-    upper: i64,
+    lower: i64, // slope 1600/lower just outside left boundary
+    upper: i64, // slope upper/1600 just outside right boundary
```
ebblake: I plan to revise this to track scale as part of the struct rather than hard-coded

src/year2019/day19.rs
```diff
@@ -2,26 +2,65 @@
...
 }
 
 pub fn parse(input: &str) -> Input {
     let code: Vec<_> = input.iter_signed().collect();
     let mut lower = 0;
     let mut upper = 0;
+    let mut scale = 50;
```
ebblake: Starting at a smaller scale allows for a faster parse, as long as the scale works. Scale=5 fails on [Eric's official sample](https://github.com/topaz/aoc2019-intcode/blob/master/samples/day19.intcode) but scale=10 should work as the starting point. Testing that now before revising the patch.

src/year2019/day19.rs
```diff
@@ -70,7 +110,7 @@ pub fn part2(input: &Input) -> i64 {
 
 /// Quick check with some false positives but no false negatives.
 fn precheck(input: &Input, x: i64, y: i64) -> bool {
-    50 * y > input.upper * x && 50 * x > input.lower * y
+    1600 * y > input.upper * x && 1600 * x > input.lower * y
```
ebblake: Thus should use input.scale rather than hard-coding.

ebblake:
[Eric's official sample](https://github.com/topaz/aoc2019-intcode/blob/master/samples/day19.intcode) shows that the black box is just computing a linear inequality: `abs(c1*x*x - c2*y*y) < c3*x*y` for three constants chosen so that there are no integer hits on y=1. The slopes are irrational, but the better you can approximate the slope, the fewer times you have to call into intcode to see if rounding boundaries of the actual slopes match your estimates just outside the slope.

Starting scale at 50 requires about 100 test() to find the first estimate. Each later scale by 2 requires between 2 and 6 further probes (possibly between 0 and 4 if the loop logic is improved) to cut the error between the approximate and actual slope in half. With no pre-filtering by approximate slope, crawling the diagonals on part 1 needs around 200 test() (better than brute force 2500 probes) and part 2 walking along the diagonal until reaching the winning spot needs around 700-1000 probes. But with pre-filtering and intelligent starting points, part 1 can probe fewer than 100 times, and part 2 around 10 times.

You could also break the black box and extract the 3 constants to solve this exactly with zero lines of intcode run.

ebblake:
The following hack (post-patch) makes it easier to see the effects of different scale and part 2 starting values:
```
diff --git c/src/year2019/day19.rs i/src/year2019/day19.rs
index 5d09a42..0a53508 100644
--- c/src/year2019/day19.rs
+++ i/src/year2019/day19.rs
@@ -53,20 +53,20 @@ pub struct Input {
     // Pick an initial scale large enough to be past the discontinuities for all known inputs.
     let code: Vec<_> = input.iter_signed().collect();
-    let mut lower = 1;
-    let mut upper = 1;
-    let mut scale = 5;
+    let mut lower = 0;
+    let mut upper = 0;
+    let mut scale = 25;

     // Find approximate slope of lower and upper edges, rounding down to prevent false negatives.
     // Scale the boundary for slightly more accuracy.
-    while scale < 1024 {
+    while scale < 50 {
         scale *= 2;
         lower *= 2;
         upper *= 2;
-        while !test(&code, lower + 1, scale) {
+        while !test(&code, lower + 1, scale, 0) {
             lower += 1;
         }
-        while !test(&code, scale, upper + 1) {
+        while !test(&code, scale, upper + 1, 0) {
             upper += 1;
         }
     }
@@ -81,9 +81,9 @@ pub fn part1(input: &Input) -> i64 {

     // Scanning the remaining lines works even around the known discontinuity at y=1, by finding
     // the left and right edges if any.
-    for y in 2..50 {
-        let left = (1..50).find(|&x| precheck(input, x, y) && test(code, x, y));
-        let right = (1..50).rfind(|&x| precheck(input, x, y) && test(code, x, y));
+    for y in 0..50 {
+        let left = (0..50).find(|&x| precheck(input, x, y, 1) && test(code, x, y, 1));
+        let right = (0..50).rfind(|&x| precheck(input, x, y, 1) && test(code, x, y, 1));
         if let (Some(l), Some(r)) = (left, right) {
             result += r - l + 1;
         }
@@ -99,17 +99,18 @@ pub fn part2(input: &Input) -> i64 {
         / (input.scale * input.scale - input.lower * input.upper);
     let mut y = input.scale * x / input.lower - 99;
     let mut moved = true;
+    x = 0; y = 0;

     // Increase the right and bottom edges of our box until they are both inside the beam.
     while moved {
         moved = false;

-        while !precheck(input, x, y + 99) || !test(code, x, y + 99) {
+        while !precheck(input, x, y + 99, 2) || !test(code, x, y + 99, 2) {
             x += 1;
             moved = true;
         }

-        while !precheck(input, x + 99, y) || !test(code, x + 99, y) {
+        while !precheck(input, x + 99, y, 2) || !test(code, x + 99, y, 2) {
             y += 1;
             moved = true;
         }
@@ -119,12 +120,14 @@ pub fn part2(input: &Input) -> i64 {
 }

 /// Quick check with some false positives but no false negatives.
-fn precheck(input: &Input, x: i64, y: i64) -> bool {
+fn precheck(input: &Input, x: i64, y: i64, p: u8) -> bool {
+    println!("precheck{p} {x} {y}");
     input.scale * y > input.upper * x && input.scale * x > input.lower * y
 }

 /// Definitive but slower check.
-fn test(code: &[i64], x: i64, y: i64) -> bool {
+fn test(code: &[i64], x: i64, y: i64, p: u8) -> bool {
+    println!("test{p} {x} {y}");
     let mut computer = Computer::new(code);
     computer.input(x);
     computer.input(y);
```
when run with
```
cargo run year2019::day19 | sed 's/ .*//' | sort | uniq -c
```

With scale = 50 and x,y of 0,0 (the pre-patch condition), on my input, it produces:
```
   2717 precheck1
   1882 precheck2
     93 test0
     94 test1
    266 test2
```
With scale starting at 10 and grown past 1024, and x,y determined by geometry, it produces:
```
   2464 precheck1
     29 precheck2
     39 test0
     86 test1
     20 test2
```
Additionally, you can inspect how often a given point is being probed more than once, to decide if memoization is worthwhile. Pre-patch:
```
$ cargo run year2019::day19 | sort | uniq -c | sed 's/\([tk][012]\).*/\1/' | sort | uniq -c
...
   2009       1 precheck1
   1878       1 precheck2
     93       1 test0
     86       1 test1
    262       1 test2
    354       2 precheck1
      2       2 precheck2
      4       2 test1
      2       2 test2
```
Post-patch:
```
   1966       1 precheck1
     25       1 precheck2
     39       1 test0
     78       1 test1
     16       1 test2
    249       2 precheck1
      2       2 precheck2
      4       2 test1
      2       2 test2
```
So no point is being probed 3 times (good), and the number of points probed twice is fairly small (unchanged by my patch), and memoization is going to cost more memory than time saved. Further improvements could be made by enhancing part 1 to more closely scan the two lines (advancing left and right independently by 0, 1, or 2) rather than doing a binary search on EACH line and therefore having a several wasted calls to test() when precheck() returns true for a point definitively in the middle of the line, rather than on the 2-cell border that needs actual probing.

src/year2019/day19.rs
ebblake: The latest patch starts at scale=5 outside the loop, but doubles it before testing, so it uses the safe scale=10 on Eric's input.

ebblake:
I also tried scale>2048 - it cost 4 more test in parse but saved 7 in part2. That is starting to reach the point of diminishing returns

src/year2019/day19.rs
```diff
@@ -1,40 +1,88 @@
...
+//!           1111111
+//!            11111111
+//!             111SxxA11  <--- (x1, y1)
+//!              11xxxx1111
+//!               1xxxx111111
+//!      (x2, y2)> Bxxx11111111
+//!                 1111111111111
+//! ```
+//!
+//! For a 100x100 box, we have a system of equations m1=upper/scale≈y1/x1, x1=x2+99,
```
maneatingape: Alternatively if (x,y) is the top left corner of the box, then there are only 2 equations:
s = scale, u = upper, l = lower
sy = u(x + 99)
sx = l(y + 99)

In matrix form:
[u -s][x] = [-99u]
[s -l][y] = [99l]

Inverting the 2 x 2 matrix gives:
```
    let determinant = input.scale * input.scale - input.lower * input.upper;
    let mut x = 99 * (input.lower * input.upper + input.lower * input.scale) / determinant;
    let mut y = 99 * (input.lower * input.upper + input.upper * input.scale) / determinant;
```
A little verbose but clearer (and gives exactly the same results)

src/year2019/day19.rs
```diff
@@ -1,40 +1,88 @@
...
+    // The origin is always set, but no other point occurs on that row or column.
+    let mut result = 1;
 
-    // The beam is continuous so we only need to find the left and right edges.
-    for y in 0..50 {
-        let left = (0..50).find(|&x| inside(input, x, y));
-        let right = (0..50).rfind(|&x| inside(input, x, y));
+    // Scanning the remaining lines works even around the known discontinuity at y=1, by finding
+    // the left and right edges if any.
+    for y in 2..50 {
```
maneatingape: `for y in 1..50` makes no assumptions about the discontinuity and benchmark difference is in the noise.

src/year2019/day19.rs
```diff
@@ -1,40 +1,88 @@
...
+    let mut scale = 5;
 
-    // Find slope of lower and upper edges, rounding down to prevent false negatives.
-    while !test(&code, lower + 1, 50) {
-        lower += 1;
-    }
-    while !test(&code, 50, upper + 1) {
-        upper += 1;
+    // Find approximate slope of lower and upper edges, rounding down to prevent false negatives.
+    // Scale the boundary for slightly more accuracy.
```
maneatingape: ❤️

maneatingape (changes requested):
Nice speedup, especially part two.

src/year2019/day19.rs
ebblake: Even better: start the .rfind search for right at left, rather than 1.  On rows with a discontinuity, this avoids scanning the entire line twice.

src/year2019/day19.rs
```diff
@@ -1,40 +1,73 @@
 //! # Tractor Beam
 //!
-//! Finds the approximate boundary of the upper and lower edges of the beam expressed as a slope.
-//! We then skip the relatively expensive intcode test if the x and y coordinates lie outside.
+//! The intcode program computes a linear inequality: returning true if an integer point lies on
+//! or between two lines through the origin, often with irrational slope. The intcode program was
```
ebblake, line 4: My input file was irrational for both lines, but I did find another input file where one of the two lines ended up with a rational slope. The solution works either way.

maneatingape (approved):
👍

PR #74, opened 2026-05-21
Use A* search for less overall work

ebblake:
## Description

The search can be steered towards the global minimum by pre-computing a heuristic of the absolute minimum possible distance that each key can contribute, then updating that heuristic with a subtraction per key visited, making it a nice lightweight O(1) heuristic.  By the time we reach the goal of zero keys remaining, the heuristic also reaches zero, so it is consistent and we do not have to worry about revisiting a node with a lower distance from an alternative path.  Then, using an O(n) heuristic for part 2 (but not part 1) got even more speed by pruning much more of the search space.

On my laptop, this improves runtimes dramtically, even if the numbers are a bit noisy.  A more direct analysis, by instrumenting the number of cache accesses and items of work processed when using my input file, shows that part 2 sees the bigger benefit:

```
                        cache.entry()  todo.push()  todo.pop() runtime
    part 1 pre-series    101494         23195        22762      5.2ms
    part 1 post-series    65546         23019        21410      4.6ms
    part 2 pre-series    120813         33238        31437     12.1ms
    part 2 post-series    33577         13826         6689      2.3ms
```

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day18.rs
+++ b/src/year2019/day18.rs
@@ -69,14 +69,18 @@ struct Door {
     needed: u32,
 }
 
+type Matrix = [[Door; 30]; 30];
+
 /// `initial` is the complete set of keys that we need to collect. Will always be binary
 /// `11111111111111111111111111` for the real input but fewer for sample data.
 ///
-/// `maze` is the adjacency matrix of distances and doors between each pair of keys and the
-/// robots' starting locations.
+/// `masks` maps the set of keys in the same quadrant, for prefiltering in part 2.
+/// `matrix` is the adjacency of distances and doors between each pair of keys and the robots'
+/// starting locations.
 struct Maze {
     initial: State,
-    maze: [[Door; 30]; 30],
+    masks: [u32; 30],
+    matrix: Matrix,
 }
 
 pub fn parse(input: &str) -> Grid<u8> {
@@ -124,9 +128,10 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
     // As a minor optimization we reuse the same `todo` and `seen` between each search.
     let default = Door { distance: u32::MAX, needed: 0 };
 
-    let mut maze = [[default; 30]; 30];
+    let mut matrix = [[default; 30]; 30];
     let mut seen = vec![usize::MAX; bytes.len()];
     let mut todo = VecDeque::new();
+    let mut masks = [0; 30];
 
     for (start, from) in found {
         todo.push_front((start, 0, 0));
@@ -141,8 +146,10 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
                 && distance > 0
             {
                 // Store the reciprocal edge weight and doors in the adjacency matrix.
-                maze[from][to] = Door { distance, needed };
-                maze[to][from] = Door { distance, needed };
+                matrix[from][to] = Door { distance, needed };
+                matrix[to][from] = Door { distance, needed };
+                masks[from] |= 1 << to;
+                masks[to] |= 1 << from;
                 // Faster to stop here and use Floyd-Warshall later.
                 continue;
             }
@@ -159,31 +166,33 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
     // Fill in the rest of the graph using the Floyd-Warshall algorithm.
     // As a slight twist we also build the list of intervening doors at the same time.
     for i in RANGE {
-        maze[i][i].distance = 0;
+        matrix[i][i].distance = 0;
     }
 
     for k in RANGE {
         for i in RANGE {
             for j in RANGE {
-                let candidate = maze[i][k].distance.saturating_add(maze[k][j].distance);
-                if maze[i][j].distance > candidate {
-                    maze[i][j].distance = candidate;
+                let candidate = matrix[i][k].distance.saturating_add(matrix[k][j].distance);
+                if matrix[i][j].distance > candidate {
+                    matrix[i][j].distance = candidate;
                     // `(1 << k)` is a crucial optimization. By treating intermediate keys like
                     // doors we speed things up by a factor of 30.
-                    maze[i][j].needed = maze[i][k].needed | (1 << k) | maze[k][j].needed;
+                    matrix[i][j].needed = matrix[i][k].needed | (1 << k) | matrix[k][j].needed;
+                    masks[i] |= 1 << j;
+                    masks[j] |= 1 << i;
                 }
             }
         }
     }
 
-    Maze { initial, maze }
+    Maze { initial, masks, matrix }
 }
 
 fn explore(width: usize, bytes: &[u8]) -> u32 {
     let mut todo = MinHeap::with_capacity(5_000);
     let mut cache = FastMap::with_capacity(5_000);
 
-    let Maze { initial, maze } = parse_maze(width, bytes);
+    let Maze { initial, masks, matrix } = parse_maze(width, bytes);
     todo.push(0, initial);
 
     while let Some((total, State { position, remaining })) = todo.pop() {
@@ -193,14 +202,22 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
             return total;
         }
 
+        // Avoid next-neighbor checks if this state was visited in the meantime by a better path.
+        if let Some(&best) = cache.get(&State { position, remaining })
+            && total > best
+        {
+            continue;
+        }
+
         // The set of robots is stored as bits in a `u32` shifted by the index of the location.
         for from in position.biterator() {
             // The set of keys still needed is also stored as bits in a `u32` similarly to robots.
-            for to in remaining.biterator() {
-                let Door { distance, needed } = maze[from][to];
+            // Filter the list of destinations to keys in the same quadrant.
+            for to in (remaining & masks[from]).biterator() {
+                let Door { distance, needed } = matrix[from][to];
 
-                // u32::MAX indicates that two nodes are not connected. Only possible in part two.
-                if distance != u32::MAX && remaining & needed == 0 {
+                // Don't move to a key that still has unmet dependencies.
+                if remaining & needed == 0 {
                     let next_total = total + distance;
                     let from_mask = 1 << from;
                     let to_mask = 1 << to;
--- a/src/year2019/day18.rs
+++ b/src/year2019/day18.rs
@@ -7,10 +7,15 @@
 //!
 //! We first find the distance between every pair of keys then run the
 //! [A* algorithm](https://en.wikipedia.org/wiki/A*_search_algorithm) to find the
-//! shortest path that visits every node in the graph. The heuristic is the sum of the minimum
-//! possible distances to reach every remaining key, which helps the search focus on the
-//! overall minimum rather than the nearest key, for fewer nodes explored. Since the heuristic
-//! is consistent, no node will ever ever have its score reduced after the first visit.
+//! shortest path that visits every node in the graph. One heuristic is a constant-time query
+//! of the sum of the minimum path length out of all remaining keys (updated by subtraction
+//! as a key is visited), which works well for part one when there is only one robot that must
+//! visit every remaining node, but underestimates for part two. Another heuristic is a cacheable
+//! linear-time query of the maximum distance each robot must travel to reach the furthest remaining
+//! key. This latter query is expensive enough that it penalizes part one, but its improved accuracy
+//! prunes more states in part two than the extra time spent on computing the heuristic. Both
+//! heuristics reach zero at the goal, and are consistent, meaning they never overestimate and no
+//! state will lower its score after the initial visit.
 //!
 //! The maze is also constructed in such a way to make our life easier:
 //! * There is only ever one possible path to each key. We do not need to consider
@@ -78,7 +83,7 @@ type Matrix = [[Door; 30]; 30];
 /// `11111111111111111111111111` for the real input but fewer for sample data.
 ///
 /// `masks` maps the set of keys in the same quadrant, for prefiltering in part 2.
-/// `minimum` is the smallest distance from a key to any of its neighbors, for the A* heuristic.
+/// `minimum` is the smallest distance from a key to any of its neighbors, for the part1 heuristic.
 /// `matrix` is the adjacency of distances and doors between each pair of keys and the robots'
 /// starting locations.
 struct Maze {
@@ -93,7 +98,8 @@ pub fn parse(input: &str) -> Grid<u8> {
 }
 
 pub fn part1(input: &Grid<u8>) -> u32 {
-    explore(input.width as usize, &input.bytes)
+    // Select the O(1) A* heuristic, since there is only one robot visiting all keys.
+    explore::<false>(input.width as usize, &input.bytes)
 }
 
 pub fn part2(input: &Grid<u8>) -> u32 {
@@ -108,7 +114,8 @@ pub fn part2(input: &Grid<u8>) -> u32 {
     patch("###", 0);
     patch("@#@", 1);
 
-    explore(input.width as usize, &modified)
+    // Select the O(n) A* heuristic, since each robot vists about one-fourth of the keys.
+    explore::<true>(input.width as usize, &modified)
 }
 
 fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
@@ -196,16 +203,21 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
     Maze { initial, masks, minimum, matrix }
 }
 
-fn explore(width: usize, bytes: &[u8]) -> u32 {
+// Same algorithm, but specialized on which heuristic to use.
+fn explore<const PART_TWO: bool>(width: usize, bytes: &[u8]) -> u32 {
     let mut todo = MinHeap::with_capacity(5_000);
-    let mut cache = FastMap::with_capacity(5_000);
+    let mut state_cache = FastMap::with_capacity(5_000);
+    let mut heur_cache = FastMap::with_capacity(5_000);
 
     let Maze { initial, masks, minimum, matrix } = parse_maze(width, bytes);
-    let heuristic: u32 = minimum.iter().filter(|&min| *min < u32::MAX).sum();
-    todo.push(heuristic, (initial, heuristic));
+    let heur = if PART_TWO {
+        heuristic(initial, &masks, &matrix, &mut heur_cache)
+    } else {
+        minimum.iter().filter(|&min| *min < u32::MAX).sum()
+    };
+    todo.push(heur, (initial, 0));
 
-    while let Some((guess, (State { position, remaining }, heuristic))) = todo.pop() {
-        let total = guess - heuristic;
+    while let Some((guess, (State { position, remaining }, total))) = todo.pop() {
         // Finish immediately if no keys left.
         // Since we're using A* with a consistent heuristic this will always be the optimal solution.
         if remaining == 0 {
@@ -213,7 +225,7 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
         }
 
         // Avoid next-neighbor checks if this state was visited in the meantime by a better path.
-        if let Some(&best) = cache.get(&State { position, remaining })
+        if let Some(&best) = state_cache.get(&State { position, remaining })
             && total > best
         {
             continue;
@@ -237,12 +249,16 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
                     };
 
                     // Memoize previously seen states to eliminate suboptimal states right away.
-                    let best = cache.entry(next_state).or_insert(u32::MAX);
+                    let best = state_cache.entry(next_state).or_insert(u32::MAX);
                     if next_total < *best {
                         *best = next_total;
-                        let next_heuristic = heuristic - minimum[to];
-                        let next_guess = next_total + next_heuristic;
-                        todo.push(next_guess, (next_state, next_heuristic));
+                        let next_heur = if PART_TWO {
+                            heuristic(next_state, &masks, &matrix, &mut heur_cache)
+                        } else {
+                            guess - total - minimum[to]
+                        };
+                        let next_guess = next_total + next_heur;
+                        todo.push(next_guess, (next_state, next_total));
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


ebblake:
`just` gave me a weird warning that seemed unrelated to my original patch, hence my s/maze/matrix/ to silence it:
```
warning: field name starts with the struct's name
  --> src/year2019/day18.rs:80:5
   |
80 |     maze: [[Door; 30]; 30],
   |     ^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names
   = note: requested on the command line with `-W clippy::struct-field-names`
```

ebblake:
Here's an alternative O(n) heuristic. It made part1 almost twice as long (the number of keys reachable from one bot is high enough that a lot of time is being spent on computing the max leg remaining), but part2 noticeably faster (with four bots, the effort to find max leg is spread out more evenly, and max leg rather than sum of min legs steers better, even though it requires more effort to compute). Maybe a hybrid is worthwhile? (No heuristic for part1, O(n) heuristic for part 2)
```
From 84c924e844b0da501c57c7c3aafb41e46b05b25d Mon Sep 17 00:00:00 2001
From: Eric Blake <eblake@redhat.com>
Date: Thu, 21 May 2026 10:37:28 -0500
Subject: [PATCH] tmp
Content-type: text/plain

---
 src/year2019/day18.rs | 46 ++++++++++++++++++++++++++-----------------
 1 file changed, 28 insertions(+), 18 deletions(-)

diff --git a/src/year2019/day18.rs b/src/year2019/day18.rs
index 606c2f7..536b932 100644
--- a/src/year2019/day18.rs
+++ b/src/year2019/day18.rs
@@ -72,16 +72,16 @@ struct Door {
     needed: u32,
 }

+type Matrix = [[Door; 30]; 30];
+
 /// `initial` is the complete set of keys that we need to collect. Will always be binary
 /// `11111111111111111111111111` for the real input but fewer for sample data.
 ///
 /// `matrix` is the adjacency of distances and doors between each pair of keys and the robots
 /// starting locations.
-/// `minimum` is the smallest distance from a key to any of its neighbors, for the A* heuristic.
 struct Maze {
     initial: State,
-    matrix: [[Door; 30]; 30],
-    minimum: [u32; 26],
+    matrix: Matrix,
 }

 pub fn parse(input: &str) -> Grid<u8> {
@@ -181,25 +181,19 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
         }
     }

-    let mut minimum = [0; 26];
-    for i in initial.remaining.biterator() {
-        minimum[i] =
-            matrix[i].iter().map(|d| d.distance).filter(|&dist| dist > 0).min().unwrap_or(0);
-    }
-
-    Maze { initial, matrix, minimum }
+    Maze { initial, matrix }
 }

 fn explore(width: usize, bytes: &[u8]) -> u32 {
     let mut todo = MinHeap::with_capacity(5_000);
     let mut cache = FastMap::with_capacity(5_000);

-    let Maze { initial, matrix, minimum } = parse_maze(width, bytes);
-    let heuristic = minimum.iter().sum();
-    todo.push(heuristic, (initial, heuristic));
+    let Maze { initial, matrix } = parse_maze(width, bytes);
+    let heur = heuristic(initial, &matrix);
+    todo.push(heur, (initial, heur));

-    while let Some((guess, (State { position, remaining }, heuristic))) = todo.pop() {
-        let total = guess - heuristic;
+    while let Some((guess, (State { position, remaining }, heur))) = todo.pop() {
+        let total = guess - heur;
         // Finish immediately if no keys left.
         // Since we're using A* with a consistent heuristic this will always be the optimal solution.
         if remaining == 0 {
@@ -220,14 +214,14 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
                         position: position ^ from_mask ^ to_mask,
                         remaining: remaining ^ to_mask,
                     };
-                    let next_heuristic = heuristic - minimum[to];
-                    let next_guess = total + distance + next_heuristic;
+                    let next_heur = heuristic(next_state, &matrix);
+                    let next_guess = total + distance + next_heur;

                     // Memoize previously seen states to eliminate suboptimal states right away.
                     let best = cache.entry(next_state).or_insert(u32::MAX);
                     if next_guess < *best {
                         *best = next_guess;
-                        todo.push(next_guess, (next_state, next_heuristic));
+                        todo.push(next_guess, (next_state, next_heur));
                     }
                 }
             }
@@ -245,3 +239,19 @@ fn is_key(b: u8) -> Option<usize> {
 fn is_door(b: u8) -> Option<usize> {
     b.is_ascii_uppercase().then(|| (b - b'A') as usize)
 }
+
+fn heuristic(state: State, matrix: &Matrix) -> u32 {
+    let mut heur = 0;
+
+    // For each robot, compute the worst-case distance it must travel to a remaining key.
+    for bot in state.position.biterator() {
+        let mut dist = 0;
+        for key in state.remaining.biterator() {
+            if matrix[bot][key].distance != u32::MAX {
+                dist = dist.max(matrix[bot][key].distance);
+            }
+        }
+        heur += dist;
+    }
+    heur
+}
-- 
2.54.0

```
with this alternative, tracing shows
```
                    cache.entry()  todo.push()  todo.pop()  runtime
part 1 pre-patch    101494         23195        22762        5.3ms
part 1 O(1)-patch    96755         23054        21528        5.0ms
part 1 O(n)-patch   110269         25376        24334       10.1ms
part 2 pre-patch    120813         31437        33238       12.3ms
part 2 O(1)-patch    99325         30438        24419       10.7ms
part 2 O(n)-patch    33625         13826         6689        8.0ms
```

ebblake:
Gemini AI suggests this approach for a hybrid, that picks the best heuristic for both parts:
```

// The const generic flag tells the compiler to specialize this function
fn explore<const IS_PART_2: bool>(width: usize, bytes: &[u8]) -> u32 {
    let mut todo = MinHeap::with_capacity(5_000);
    let mut cache = FastMap::with_capacity(5_000);
    let Maze { initial, matrix, minimum } = parse_maze(width, bytes);

    // Compute initial heuristic based on the compile-time flag
    let heur = if IS_PART_2 {
        heuristic_part2(initial, &matrix)
    } else {
        minimum.iter().sum()
    };

    todo.push(heur, (initial, heur));

    while let Some((guess, (state, heur))) = todo.pop() {
        // ... (hot loop logic) ...

        let next_heur = if IS_PART_2 {
            heuristic_part2(next_state, &matrix)
        } else {
            heur - minimum[to] // Fast O(1) step subtraction
        };
    }
    // ...
}
```

src/year2019/day18.rs
```diff
@@ -197,23 +210,24 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
         for from in position.biterator() {
             // The set of keys still needed is also stored as bits in a `u32` similarly to robots.
             for to in remaining.biterator() {
-                let Door { distance, needed } = maze[from][to];
+                let Door { distance, needed } = matrix[from][to];
 
                 // u32::MAX indicates that two nodes are not connected. Only possible in part two.
                 if distance != u32::MAX && remaining & needed == 0 {
```
ebblake: Instead of having to check distance != u32::MAX, would it be any more efficient to prefilter things via (remaining & quadrant_mask[from]).biterator() to reduce the biterator to just the neighbors that the robot can reach?

ebblake: Yep, it was indeed. I've resubmitted, now as a series that gets part 2 down below 33% of its original runtime.

src/year2019/day18.rs
```diff
@@ -159,61 +185,86 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
+                    };
+                    let next_guess = total + distance + next_heur;
 
                     // Memoize previously seen states to eliminate suboptimal states right away.
                     let best = cache.entry(next_state).or_insert(u32::MAX);
-                    if next_total < *best {
-                        *best = next_total;
-                        todo.push(next_total, next_state);
+                    if next_guess < *best {
+                        *best = next_guess;
```
ebblake: Better is to cache the distance, not distance+heuristic. I'm revising accordingly.

src/year2019/day18.rs
```diff
@@ -159,61 +185,86 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
 
-    let Maze { initial, maze } = parse_maze(width, bytes);
-    todo.push(0, initial);
+    let Maze { initial, masks, minimum, matrix } = parse_maze(width, bytes);
+    let heur = if PART_TWO { heuristic(initial, &masks, &matrix) } else { minimum.iter().sum() };
+    todo.push(heur, (initial, heur));
 
-    while let Some((total, State { position, remaining })) = todo.pop() {
+    while let Some((guess, (State { position, remaining }, heur))) = todo.pop() {
+        let total = guess - heur;
```
ebblake: I added some extra tracing. Even without A*, Dijkstra queues nodes at one depth, and then revisits them from another cheaper path. The node is still popped twice, and on the second pop, we end up going through all the nearest neighbor checks only to find nothing further to do. Adding a post-pop check against the cache might reduce some cache pressure for even more speed. Testing that now.

ebblake: Yep, checking the cache up front saved more time (6837 nodes out of 21410 revisited in the queue for part 1; doing 21k cache checks up front saves more than 30k accesses later on).  I folded that improvement into the latest round of patches.

ebblake:
I'm still investigating if a dynamic programming approach similar to Held-Karp can outperform A*. One benefit of the dynamic approach is that you can solve both parts at once. With A*, your cache is on (keys,positions)->distance, which is two separate caches for the two parts. But with dynamic programming, the cache is (keys,last key)->(distance,positions) which can cram both parts into the same cache for less traversal overhead. The dynamic approach is up to O(n^2) per set visited (for each reachable key in a set, find the minimum distance when appending that key to any of the other reachable keys of the subset - in general n is much smaller than 26 because of the non-viable keys with a nonzero need), but I don't yet have a feel for how many set visits can be pruned from the reachability front to avoid a full-blown 2^26 set visits.

ebblake:
> I'm still investigating if a dynamic programming approach similar to Held-Karp can outperform A*.

Nope - although I did get a dynamic search working, it visits more sets+positions (31272 on my input - but still much better than a full 2^26 sets) than a directed A* search. At this point, I'm happy with the current state of the patch as being the fastest solution I can come up with.

src/year2019/day18.rs
```diff
@@ -231,3 +295,18 @@ fn is_key(b: u8) -> Option<usize> {
 fn is_door(b: u8) -> Option<usize> {
     b.is_ascii_uppercase().then(|| (b - b'A') as usize)
 }
+
+// Compute part two heuristic of the sum of the furthest key remaining per robot. For part one,
+// rely on the faster but weaker O(1) tracking of the sum of all minimum legs.
+fn heuristic(state: State, masks: &[u32], matrix: &Matrix) -> u32 {
```
maneatingape: Adding a cache sped things up about 10-15%, e.g.

```
// Compute part two heuristic of the sum of the furthest key remaining per robot. For part one,
// rely on the faster but weaker O(1) tracking of the sum of all minimum legs.
fn heuristic(state: State, masks: &[u32], matrix: &Matrix, cache: &mut FastMap<(usize, u32), u32>) -> u32 {
    let mut heur = 0;

    for bot in state.position.biterator() {
        let reachable = state.remaining & masks[bot];

        let dist = *cache.entry((bot, reachable)).or_insert_with(|| {
            reachable.biterator().map(|key| matrix[bot][key].distance).max().unwrap_or(0)
        });

        heur += dist;
    }
    heur
}
```

src/year2019/day18.rs
```diff
@@ -159,48 +190,75 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
+        {
+            continue;
+        }
+
         // The set of robots is stored as bits in a `u32` shifted by the index of the location.
         for from in position.biterator() {
             // The set of keys still needed is also stored as bits in a `u32` similarly to robots.
-            for to in remaining.biterator() {
-                let Door { distance, needed } = maze[from][to];
+            // Filter the list of destinations to keys in the same quadrant.
```
maneatingape, line 237: ❤️

src/year2019/day18.rs
```diff
@@ -213,7 +271,13 @@ fn explore(width: usize, bytes: &[u8]) -> u32 {
                     let best = cache.entry(next_state).or_insert(u32::MAX);
                     if next_total < *best {
                         *best = next_total;
-                        todo.push(next_total, next_state);
+                        let next_heur = if PART_TWO {
+                            heuristic(next_state, &masks, &matrix)
+                        } else {
+                            heur - minimum[to]
```
maneatingape: Changing this to 0 for part one (i.e. no heuristic at all) was either the same or slightly faster when I benchmarked.

src/year2019/day18.rs
```diff
@@ -159,48 +190,75 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
     let mut cache = FastMap::with_capacity(5_000);
 
-    let Maze { initial, maze } = parse_maze(width, bytes);
-    todo.push(0, initial);
+    let Maze { initial, masks, minimum, matrix } = parse_maze(width, bytes);
+    let heur = if PART_TWO { heuristic(initial, &masks, &matrix) } else { minimum.iter().sum() };
+    todo.push(heur, (initial, heur));
 
-    while let Some((total, State { position, remaining })) = todo.pop() {
+    while let Some((guess, (State { position, remaining }, heur))) = todo.pop() {
```
maneatingape: Simplify by just passing the total directly and ignoring heuristic (as it's only needed for heap ordering) e.g.

`while let Some((_, (State { position, remaining }, total))) = todo.pop() {`

maneatingape (changes requested):
Nice heuristic for part two.

src/year2019/day18.rs
ebblake: Odd. I did see a (slight) speedup on my end, and instrumenting the number of cache pops definitely saw a (small) improvement. But maybe other input files are less sensitive to the gains. Would you like to (temporarily) apply the patch in my other comment to count `cache.entry()  todo.push()  todo.pop()` with the heuristic vs with 0?

src/year2019/day18.rs
ebblake: Odd. Making just this change (which should have no semantic impact) actually slowed down part1() (on my machine, it dropped from 4.6ms to 5.1ms).  At any rate, the slowdown I observe with this patch may be why you made the suggestion to not have a heuristic for part 1?
```
diff --git i/src/year2019/day18.rs w/src/year2019/day18.rs
index 2c94780..0452888 100644
--- i/src/year2019/day18.rs
+++ w/src/year2019/day18.rs
@@ -233,10 +233,9 @@ fn explore<const PART_TWO: bool>(width: usize, bytes: &[u8]) -> u32 {

     let Maze { initial, masks, minimum, matrix } = parse_maze(width, bytes);
     let heur = if PART_TWO { heuristic(initial, &masks, &matrix) } else { minimum.iter().sum() };
-    todo.push(heur, (initial, heur));
+    todo.push(heur, (initial, 0));

-    while let Some((guess, (State { position, remaining }, heur))) = todo.pop() {
-        let total = guess - heur;
+    while let Some((_, (State { position, remaining }, total))) = todo.pop() {
         // Finish immediately if no keys left.
         // Since we're using A* with a consistent heuristic this will always be the optimal solution.
         if remaining == 0 {
@@ -277,7 +276,7 @@ fn explore<const PART_TWO: bool>(width: usize, bytes: &[u8]) -> u32 {
                             heur - minimum[to]
                         };
                         let next_guess = next_total + next_heur;
-                        todo.push(next_guess, (next_state, next_heur));
+                        todo.push(next_guess, (next_state, next_total));
                     }
                 }
             }
```

ebblake: Aha - I see why. Without that patch, the part1() computation of next_heur is `heur - minimum[to]` where heur was passed through the loop. But with that patch, heur remains the unchanging value from outside of the loop, at which point the part1 heuristic is computed incorrectly and the priority queue is no longer accurate, at which point replacing the part1 heuristic with constant 0 is no worse.

Despite how nice it normally is for Rust to let you rebind a variable, this is one case where I missed having a compiler that would notify me about the subtle change in semantics due to shadowing an earlier declaration when the rebind gates a new scope.

src/year2019/day18.rs
ebblake: Yep - cuts my part2 from 2.9ms to 2.3ms.  Folding that in.

maneatingape:
Tested on 3 inputs, mine (A) and two others B and C.

Counted the number of loop in the search, e.g.

```
diff --git forkSrcPrefix/src/year2019/day18.rs forkDstPrefix/src/year2019/day18.rs
index ffe7ccfc9993d3f8ca7aff145ef37207e6e74655..e8487b91e813bbc701f332e05af6635b3ddb077c 100644
--- forkSrcPrefix/src/year2019/day18.rs
+++ forkDstPrefix/src/year2019/day18.rs
@@ -241,10 +241,13 @@ fn explore<const PART_TWO: bool>(width: usize, bytes: &[u8]) -> u32 {
     };
     todo.push(heur, (initial, 0));
 
+    let mut count = 0;
+
     while let Some((guess, (State { position, remaining }, total))) = todo.pop() {
         // Finish immediately if no keys left.
         // Since we're using A* with a consistent heuristic this will always be the optimal solution.
         if remaining == 0 {
+            println!("count {count}");
             return total;
         }
 
@@ -255,6 +258,8 @@ fn explore<const PART_TWO: bool>(width: usize, bytes: &[u8]) -> u32 {
             continue;
         }
 
+        count += 1;
+
         // The set of robots is stored as bits in a `u32` shifted by the index of the location.
         for from in position.biterator() {
             // The set of keys still needed is also stored as bits in a `u32` similarly to robots.
```

| Input        |    A |    B |    C |
|--------------|-----:|-----:|-----:|
| Part 1 pre   | 2256 | 3056 | 5477 |
| Part 1 post  | 2243 | 3056 | 5043 |
| Part 2 pre   | 3056 | 5041 | 9150 |
| Part 2 post  | 1008 | 1908 | 1343 |

For part 2 a clear 66% reduction across the board. For part one on 2 inputs essentially the same, and about 10% on the third. This likely explains the differences in our benchmarking.

ebblake:
> Tested on 3 inputs, mine (A) and two others B and C.
> 
> Counted the number of loop in the search, e.g.

> For part 2 a clear 66% reduction across the board. For part one on 2 inputs essentially the same, and about 10% on the third. This likely explains the differences in our benchmarking.

For the inputs I have access to, mine (D) and two others (E and F), I see:

| Input | D | E | F |
| -- | -- | -- | -- |
| part 1 pre | 22761 | 12085 | 3897 |
| part 1 post | 14572 | 8475 | 2624 |
| part 2 pre | 31436 | 21849 | 3485 |
| part 2 post | 6679 | 7087 | 1661 |

Definitely some variability in the level of effort needed!

src/year2019/day18.rs
```diff
@@ -159,48 +190,80 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
                     // `(1 << k)` is a crucial optimization. By treating intermediate keys like
                     // doors we speed things up by a factor of 30.
-                    maze[i][j].needed = maze[i][k].needed | (1 << k) | maze[k][j].needed;
+                    matrix[i][j].needed = matrix[i][k].needed | (1 << k) | matrix[k][j].needed;
                 }
             }
         }
     }
 
-    Maze { initial, maze }
```
maneatingape, line 179: I think we can simplify the construction of `minimum` and `masks` to a single block here and remove the quadrant logic.

Something like:

```
    let mut minimum = [0; 30];
    let mut masks = [0; 30];

    // Or `i`, `j`
    for from in RANGE {
         for to in RANGE {
               let distance = matrix[from][to].distance;
             if distance > 0 && distance < u32::MAX {
                 distance[from] = distance[from].min(distance);
                 masks[from] |= 1 << to;
             }
         }
     }
```

src/year2019/day18.rs
```diff
@@ -159,48 +190,80 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
+    let mut state_cache = FastMap::with_capacity(5_000);
+    let mut heur_cache = FastMap::with_capacity(5_000);
 
-    let Maze { initial, maze } = parse_maze(width, bytes);
-    todo.push(0, initial);
+    let Maze { initial, masks, minimum, matrix } = parse_maze(width, bytes);
+    let heur = if PART_TWO {
+        heuristic(initial, &masks, &matrix, &mut heur_cache)
+    } else {
+        minimum.iter().sum()
```
maneatingape: Change to minimum[..27].iter().sum() to avoid the minimums set to u32::MAX

maneatingape (changes requested):

src/year2019/day18.rs
ebblake: No need for a separate loop - I just crammed in the minimum computation alongside the initial distance setups, and the masks there and also in the Floyd-Warshall pass (the Floyd-Warshall pass cannot create any new minimums, since we have no negative edges). But yes, it does look a bit cleaner without explicit quadrant tracking.

src/year2019/day18.rs
ebblake: That won't avoid minimums at u32::MAX in the testsuite where there are fewer than 26 keys. But a .filter() isn't too bad.

src/year2019/day18.rs
```diff
@@ -141,8 +159,12 @@ fn parse_maze(width: usize, bytes: &[u8]) -> Maze {
...
             {
                 // Store the reciprocal edge weight and doors in the adjacency matrix.
-                maze[from][to] = Door { distance, needed };
-                maze[to][from] = Door { distance, needed };
+                matrix[from][to] = Door { distance, needed };
+                matrix[to][from] = Door { distance, needed };
+                masks[from] |= 1 << to;
+                masks[to] |= 1 << from;
+                minimum[from] = minimum[from].min(distance);
+                minimum[to] = minimum[to].min(distance);
```
maneatingape, line 167: ❤️

maneatingape (approved):
👍

PR #78, opened 2026-05-28
Fix some minor math issues

ebblake:
## Description

I spotted several issues worth cleaning up while reading the day 16 solution. First, the binomial expansion is not n^k/k!, but rather n!/k!/(n-k)!. Second, during the fast two-thirds of part 1, a single range of the prefix sum array will be positive, so there is no need to use abs(). Finally, during part 2, the offset encoded at the start of the string must occur no later than 8 positions before the end of the string.

## Type of change

- [ ] Performance improvement
- [X] Bug fix (docs)
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`


maneatingape (approved):
Good catch.

PR #79, opened 2026-05-29
Speed up search for solution

ebblake:
## Description

Rather than iterating over all 100 possible nouns, we can observe that the constant a is much larger than b. In fact, b appears to be 1 in every valid input file (since verb is never multiplied, and is used late enough in the chain that later steps do not double its impact); however, b is larger than 1 in the hand-written
tests/year2019/day02.rs.

Of course, this problem is already fast (under 1us on my machine), but this changes part2 from 100ns to 5ns.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day02.rs
+++ b/src/year2019/day02.rs
@@ -6,8 +6,8 @@
 //!
 //! We can isolate the value of the constants a, b, c in order to speed up subsequent calculations.
 //!
-//! Since the equation is linear in noun and verb, we can efficiently solve part two by iterating
-//! over possible noun values and computing the corresponding verb directly.
+//! Since the equation is linear in noun and verb, and since a is much larger than b, we can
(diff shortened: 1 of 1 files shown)


ebblake:
If it were not for the hand-written test using a b > 1, the code could go even faster. In all actual input files, verb is never multiplied, and once verb is added into the chain near the very end, the chain is never added to itself. So in all actual input files, b==1, and you could speed up parse() by 33% by skipping the `let b = check(&code, 0 1);` phase. But that would require tweaking the unit tests as well, which I wasn't sure whether you wanted to go that far.

maneatingape (commented):
I don't think that the minimal time saving that risk introducing a bug is worth the tradeoff.

PR #81, opened 2026-06-01
Larger grid and faster startup for day 17

ebblake:
## Description

Patch #23 increased the size of EXTRA storage used by Intcode for day 17, which expands a run-length-encoded layout into an actual grid in the memory after the program. But in searching through the megathreads, I found at least one input that produced an 85x61 grid, requiring more than 5000 extra slots, and much larger than my own 51x39 grid.

The part 2 machine starts with output identical to part 1, and startup
is a rather time-consuming chunk of the overall Intcode program (the
program first decompresses run-length-encoded data into the grid that
will subsequently be used by the rest of the program, before its first
output).  Moving part 2 computation into parse lets the two parts
share the overhead of just one program startup.
    
On my laptop, this changes parse()+part2() from 0.7+1.4ms to now being
1.4ms in parse() and part2() being practically instant.

## Type of change

- [X] Performance improvement
- [X] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2019/day17.rs
+++ b/src/year2019/day17.rs
@@ -14,6 +14,10 @@
 //! Then we look for three patterns that can be repeated in any order to form the whole path.
 //! Without loss of generality the first pattern anchored at the start is always `A`,
 //! the next `B` and the last `C`.
+//!
+//! A good chunk of the Intcode runtime is spent on unpacking compressed memory into the grid
+//! that is first displayed to the user; this effort is the same between both parts. Running the
+//! entire solution in parse thus reduces the overall runtime.
 use super::intcode::*;
 use crate::util::hash::*;
 use crate::util::parse::*;
@@ -23,10 +27,10 @@ use std::iter::once;
 use std::ops::ControlFlow;
 
 pub struct Input {
-    code: Vec<i64>,
     scaffold: FastSet<Point>,
     position: Point,
     direction: Point,
+    score: i64,
 }
 
 struct Movement<'a> {
@@ -36,7 +40,9 @@ struct Movement<'a> {
 
 /// The camera output points from left to right, top to bottom.
 pub fn parse(input: &str) -> Input {
-    let code: Vec<_> = input.iter_signed().collect();
+    // Only run the part two program; its initial output matches part one, and avoiding the
+    // startup costs of a second machine results in a faster solution.
+    let code: Vec<_> = once(2).chain(input.iter_signed().skip(1)).collect();
     let mut computer = Computer::new(&code);
 
     let mut x = 0;
@@ -69,20 +75,10 @@ pub fn parse(input: &str) -> Input {
         x += 1;
     }
 
-    Input { code, scaffold, position, direction }
-}
+    let mut input = Input { scaffold, position, direction, score: 0 };
 
-pub fn part1(input: &Input) -> i32 {
-    input
-        .scaffold
-        .iter()
-        .filter(|&point| ORTHOGONAL.iter().all(|&delta| input.scaffold.contains(&(*point + delta))))
-        .map(|point| point.x * point.y)
-        .sum()
-}
-
-pub fn part2(input: &Input) -> i64 {
-    let path = build_path(input);
+    // With the scaffold now available, construct the compressed path.
+    let path = build_path(&input);
     let mut movement = Movement { routine: String::new(), functions: [None; 3] };
 
     let _unused = compress(&path, &mut movement);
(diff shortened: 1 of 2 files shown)


ebblake:
Of course, since a larger grid also implies a proportionately longer time to crawl such a grid, that particular input file that I just encountered will run slower than inputs that produce a smaller grid. It is also possible to break the black box and solve this problem with less intcode.  Part 2 can be solved with no additional Intcode beyond part 1 if you compute your own score by summing a function of x, y, and tiles cleaned so far for every position visited along the path. But that really defeats the purpose of the puzzle, since getting at the output sum without even computing the compression sequence feels kind of lame.  Part 1 can be solved with less Intcode if you are willing to expand the run-length-encoded data in the input file into your own grid and then warp the program counter past the point where it does its (slower) decompression.  Or combine both aspects for a solution with zero Intcode executed, because you solve everything from the scraped memory (both the RLE expansion of part 1 and the scoring from path tracing for part 2).

src/year2019/day17.rs
```diff
@@ -14,6 +14,10 @@
 //! Then we look for three patterns that can be repeated in any order to form the whole path.
 //! Without loss of generality the first pattern anchored at the start is always `A`,
 //! the next `B` and the last `C`.
+//!
+//! A good chunk of the Intcode runtime is spent on unpacking compressed memory into the grid
```
maneatingape, line 18: Nice observation!

src/year2019/day17.rs
```diff
@@ -23,10 +27,10 @@ use std::iter::once;
 use std::ops::ControlFlow;
 
 pub struct Input {
```
maneatingape, line 29: We can go even further and simplify this to just `type Input = (FastSet<Point>, i64);`

src/year2019/day17.rs
```diff
@@ -36,7 +40,9 @@ struct Movement<'a> {
 
 /// The camera output points from left to right, top to bottom.
 pub fn parse(input: &str) -> Input {
-    let code: Vec<_> = input.iter_signed().collect();
+    // Only run the part two program; its initial output matches part one, and avoiding the
+    // startup costs of a second machine results in a faster solution.
+    let code: Vec<_> = once(2).chain(input.iter_signed().skip(1)).collect();
```
maneatingape, line 45: ❤️

src/year2019/day17.rs
```diff
@@ -69,20 +75,10 @@ pub fn parse(input: &str) -> Input {
...
-        .iter()
-        .filter(|&point| ORTHOGONAL.iter().all(|&delta| input.scaffold.contains(&(*point + delta))))
-        .map(|point| point.x * point.y)
-        .sum()
-}
-
-pub fn part2(input: &Input) -> i64 {
-    let path = build_path(input);
+    // With the scaffold now available, construct the compressed path.
+    let path = build_path(&input);
```
maneatingape, line 81: Pass `scaffold`, `position` and `direction` to `build_path` as separate parameters, this will allow us to simplify the function signature to:

```
    fn build_path(scaffold: &FastSet<Point>, mut position: Point, mut direction: Point) -> String {
        let mut path = String::new();
        ...
    }
```

src/year2019/day17.rs
```diff
@@ -96,13 +92,23 @@ pub fn part2(input: &Input) -> i64 {
...
-
-    let mut computer = Computer::new(&modified);
     computer.input_ascii(&rules);
+    input.score = visit(computer);
 
-    visit(computer)
+    input
+}
+
+pub fn part1(input: &Input) -> i32 {
```
maneatingape, line 101: With the simplified `Input`, this becomes just `let (scaffold, _) = input;`

src/year2019/day17.rs
```diff
@@ -96,13 +92,23 @@ pub fn part2(input: &Input) -> i64 {
...
 
-    let mut modified = input.code.clone();
-    modified[0] = 2;
-
-    let mut computer = Computer::new(&modified);
     computer.input_ascii(&rules);
+    input.score = visit(computer);
 
-    visit(computer)
+    input
```
maneatingape, line 98: Becomes:

```
    let score = visit(computer);
    (scaffold, score)

```

maneatingape (changes requested):
Nice observation on the overlap between part 1 and 2

src/year2019/day17.rs
ebblake: Day 13 has a similar effect where startup can be shared between parts 1 and 2; but there, part 1 on my  machine is only 62us compared to part 2 at 2.8ms; which means shaving off the startup costs is in the noise, unlike day 17 where it makes an obvious difference.

src/year2019/day17.rs
ebblake: Indeed nicer. I've incorporated those changes.

maneatingape (approved):
👍

PR #82, opened 2026-06-05
Use Chinese Remainder Theorem rather than sieving

ebblake:
## Description

Sieving takes, on average, O(sum(factors)) iterations. Back in 2016 day 15, where there are only 7 primes all less than 100 (and their LCM still fits in 32 bits), this is fast. But for 2020 day 13, there are 9 primes with two over 400 (with LCM around 50 bits).  That means sieving will take around 1000 divisions.  Better is to just use CRT directly, implementing extended Euclidean division to compute coefficients, and modular multiplication to utilize u128 intermediates to avoid overflows, at which point the answer can be obtained with fewer hardware divisions.

On my laptop, this speeds up part2 runtime from 3.6us to 0.4us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day13.rs
+++ b/src/year2020/day13.rs
@@ -4,10 +4,14 @@
 //! The integers n₁, n₂, ... nₖ map to the bus ids which happen to be prime. This satisfies the
 //! requirement that the integers are [pairwise coprime](https://en.wikipedia.org/wiki/Coprime_integers#Coprimality_in_sets).
 //!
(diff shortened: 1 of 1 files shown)


maneatingape (commented):
Solution is already < 1µs, so the extra complexity of the CRT is not worth the tradeoff.

PR #83, opened 2026-06-06
Compute result byte-by-byte without splitting string

ebblake:
## Description

This puzzle is very amenable to processing one byte at a time, with our own state machine. This is lower overhead than letting split find the boundary between passengers only to then revisit the bytes of the passenger ourselves.

On my laptop, this cuts runtime from 73us (parse at 63us, part1 and part2 at 5 use each) to just 20us (all in parse).

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day06.rs
+++ b/src/year2020/day06.rs
@@ -10,22 +10,58 @@
 //! [bitwise OR](https://en.wikipedia.org/wiki/Bitwise_operation) then count the number of ones
 //! for each group using the blazing fast [`count_ones`] intrinsic.
 //!
-//! Part two is very similar, except that we use a bitwise AND instead.
+//! Part two is very similar, except that we use a bitwise AND instead. It is faster to run
+//! both parts at once during the parse.
 //!
 //! [`day 5`]: crate::year2020::day05
 //! [`count_ones`]: u32::count_ones
-pub fn parse(input: &str) -> Vec<u32> {
-    input.lines().map(|line| line.bytes().fold(0, |acc, b| acc | (1 << (b - b'a')))).collect()
+use std::iter::once;
+
+type Input = (u32, u32);
+
+pub fn parse(input: &str) -> Input {
+    let mut any = 0_u32;
+    let mut all = u32::MAX;
+    let mut passenger = 0;
+    let mut part_one = 0;
+    let mut part_two = 0;
+    let iter = input.bytes();
+    let mut check_group = false;
+
+    // End the input with a double newline, to include last group in the total.
+    for ch in iter.chain(once(b'\n')) {
+        match ch {
+            b'\n' if check_group => {
+                // A second newline ends one group and prepares for the next.
+                check_group = false;
+                part_one += any.count_ones();
+                part_two += all.count_ones();
+                any = 0;
+                all = u32::MAX;
+            }
+            b'\n' => {
+                // A single newline separates passengers within a group.
+                any |= passenger;
+                all &= passenger;
+                passenger = 0;
+                check_group = true;
+            }
+            _ => {
+                // A letter sets the next bit for the given passenger.
+                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
+                check_group = false;
+                passenger |= 1 << (ch & 31);
+            }
+        }
+    }
+
+    (part_one, part_two)
 }
 
-pub fn part1(input: &[u32]) -> u32 {
-    count_answers(input, u32::MIN, |group, passenger| group | passenger)
+pub fn part1(input: &Input) -> u32 {
+    input.0
 }
 
-pub fn part2(input: &[u32]) -> u32 {
-    count_answers(input, u32::MAX, |group, passenger| group & passenger)
-}
-
-fn count_answers(input: &[u32], initial: u32, combine: fn(u32, &u32) -> u32) -> u32 {
-    input.split(|&p| p == 0).map(|group| group.iter().fold(initial, combine).count_ones()).sum()
+pub fn part2(input: &Input) -> u32 {
+    input.1
 }
--- a/tests/year2020/day06.rs
+++ b/tests/year2020/day06.rs
@@ -15,7 +15,8 @@ a
 a
 a
 
-b";
+b
+";
 
 #[test]
 fn part1_test() {


ebblake:
Based on discussion at https://www.reddit.com/r/adventofcode/comments/1tyafwf/2020_day_6_in_review_custom_customs/

ebblake:
I noticed that the testsuite supplied sample data without the trailing newline (at least, until I patched it). But I'm assuming that in general your modus operandi is to use the file as downloaded, which includes the trailing newline. I've seen some people that try to write solutions that work whether or not the trailing newline is present, and in many cases, your parser doesn't care, but for this particular patch, my code depended on it being present.

src/year2020/day06.rs
```diff
@@ -10,22 +10,59 @@
...
+    let iter = input.bytes();
+    let mut check_group = false;
+
+    // End the input with a double newline, to include last group in the total.
+    for ch in iter.chain(once(b'\n')) {
+        match ch {
+            b'a'..=b'z' => {
+                // A letter sets the next bit for the given passenger.
+                check_group = false;
+                passenger |= 1 << (ch - b'a');
```
ebblake: Using `ch & 31` might be faster. (It's okay to use bits 1-26 instead of 0-25 for the set)

ebblake: Confirmed. Masking took me from 20.0us to 18.5us.

src/year2020/day06.rs
```diff
@@ -10,22 +10,60 @@
...
+    let iter = input.bytes();
+    let mut check_group = false;
+
+    // End the input with a double newline, to include last group in the total.
+    for ch in iter.chain(once(b'\n')) {
+        match ch {
+            b'a'..=b'z' => {
+                // A letter sets the next bit for the given passenger.
+                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
+                check_group = false;
```
ebblake: I tried to eliminate check_group by just having the \n branch check if any was 0; but that was surprisingly slower (24us instead of 20), probably because the processor couldn't pipeline it as efficiently.

maneatingape:
> I noticed that the testsuite supplied sample data without the trailing newline (at least, until I patched it). But I'm assuming that in general your modus operandi is to use the file as downloaded, which includes the trailing newline. I've seen some people that try to write solutions that work whether or not the trailing newline is present, and in many cases, your parser doesn't care, but for this particular patch, my code depended on it being present.

Yup, it's fine to add the trailing newline to the test to make it pass, since this mirrors the actual input.

src/year2020/day06.rs
```diff
@@ -10,22 +10,59 @@
...
+    let mut any = 0_u32;
+    let mut all = u32::MAX;
+    let mut passenger = 0;
+    let mut part_one = 0;
+    let mut part_two = 0;
+    let iter = input.bytes();
+    let mut check_group = false;
+
+    // End the input with a double newline, to include last group in the total.
+    for ch in iter.chain(once(b'\n')) {
```
maneatingape, line 32: Remove `chain` and replace return value with `(part_one + any.count_ones(), part_two + all.count_ones())` to include final group

src/year2020/day06.rs
```diff
@@ -10,22 +10,59 @@
...
+        match ch {
+            b'a'..=b'z' => {
```
maneatingape: Making this the default case `_` shaved some time off

src/year2020/day06.rs
maneatingape: Can you re-check with the changes? I found no difference replacing `check_group` with `if passenger == 0` guard.

maneatingape (changes requested):

src/year2020/day06.rs
ebblake: On my machine, I get 19.6us with the patch as currently online, vs. 23.9us with the following tweak to the patch:
```
diff --git i/src/year2020/day06.rs w/src/year2020/day06.rs
index 6be523f..4a6ef04 100644
--- i/src/year2020/day06.rs
+++ w/src/year2020/day06.rs
@@ -15,7 +15,6 @@
 //!
 //! [`day 5`]: crate::year2020::day05
 //! [`count_ones`]: u32::count_ones
-use std::iter::once;

 type Input = (u32, u32);

@@ -25,18 +24,11 @@ pub fn parse(input: &str) -> Input {
     let mut passenger = 0;
     let mut part_one = 0;
     let mut part_two = 0;
-    let iter = input.bytes();
     let mut check_group = false;

     // End the input with a double newline, to include last group in the total.
-    for ch in iter.chain(once(b'\n')) {
+    for ch in input.bytes() {
         match ch {
-            b'a'..=b'z' => {
-                // A letter sets the next bit for the given passenger.
-                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
-                check_group = false;
-                passenger |= 1 << (ch & 31);
-            }
             b'\n' if check_group => {
                 // A second newline ends one group and prepares for the next.
                 check_group = false;
@@ -52,7 +44,12 @@ pub fn parse(input: &str) -> Input {
                 passenger = 0;
                 check_group = true;
             }
-            _ => unreachable!(),
+            _ => {
+                // A letter sets the next bit for the given passenger.
+                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
+                check_group = false;
+                passenger |= 1 << (ch & 31);
+            }
         }
     }

```
In short, the .chain() appears to give better branch prediction than a special case at the end.

ebblake: Even this smaller patch that keeps .chain, but just rearranges the match statement arms, performs slower at 21.1us.
```
diff --git i/src/year2020/day06.rs w/src/year2020/day06.rs
index 6be523f..85669d0 100644
--- i/src/year2020/day06.rs
+++ w/src/year2020/day06.rs
@@ -31,12 +31,6 @@ pub fn parse(input: &str) -> Input {
     // End the input with a double newline, to include last group in the total.
     for ch in iter.chain(once(b'\n')) {
         match ch {
-            b'a'..=b'z' => {
-                // A letter sets the next bit for the given passenger.
-                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
-                check_group = false;
-                passenger |= 1 << (ch & 31);
-            }
             b'\n' if check_group => {
                 // A second newline ends one group and prepares for the next.
                 check_group = false;
@@ -52,6 +46,12 @@ pub fn parse(input: &str) -> Input {
                 passenger = 0;
                 check_group = true;
             }
+            b'a'..=b'z' => {
+                // A letter sets the next bit for the given passenger.
+                // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
+                check_group = false;
+                passenger |= 1 << (ch & 31);
+            }
             _ => unreachable!(),
         }
     }
```
I'm guessing that's because Rust is not doing a 256-byte jump table, but a series of conditionals; and that the conditional for letters is encountered more frequently than the conditionals for newline.

src/year2020/day06.rs
ebblake: Yes, it does look like my machine gets down to 18.0us when using a catchall for letters, so I'm folding that change in.

src/year2020/day06.rs
ebblake: On my machine, I'm consistently seeing 21.5us with check_group in place, vs. 23.9us with this patch:
```
diff --git i/src/year2020/day06.rs w/src/year2020/day06.rs
index 62e1abc..cb71d32 100644
--- i/src/year2020/day06.rs
+++ w/src/year2020/day06.rs
@@ -26,14 +26,12 @@ pub fn parse(input: &str) -> Input {
     let mut part_one = 0;
     let mut part_two = 0;
     let iter = input.bytes();
-    let mut check_group = false;

     // End the input with a double newline, to include last group in the total.
     for ch in iter.chain(once(b'\n')) {
         match ch {
-            b'\n' if check_group => {
+            b'\n' if passenger == 0 => {
                 // A second newline ends one group and prepares for the next.
-                check_group = false;
                 part_one += any.count_ones();
                 part_two += all.count_ones();
                 any = 0;
@@ -44,12 +42,10 @@ pub fn parse(input: &str) -> Input {
                 any |= passenger;
                 all &= passenger;
                 passenger = 0;
-                check_group = true;
             }
             _ => {
                 // A letter sets the next bit for the given passenger.
                 // Masking to bits 1-26 is more efficient than subtracting to bits 0-25.
-                check_group = false;
                 passenger |= 1 << (ch & 31);
             }
         }
```

I'm not sure if it is register pressure or branch prediction pollution causing the slowdown without the extra variable.

maneatingape (approved):
👍

PR #84, opened 2026-06-07
2020 day06

TerjeWiigMathisen:
## Description

byte by byte parsing, solving both parts at the same time

## Type of change

- [x] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [ ] Pull request title and commit messages are clear and informative.
- [ ] Documentation has been updated if necessary.
- [ ] Code style matches the existing code. This one is somewhat subjective, but try to "fit in" by
- [ ] Tests pass `cargo test`
- [ ] Code is formatted ``cargo fmt -- `find . -name "*.rs"` ``
- [ ] Code is linted `cargo clippy --all-targets --all-features`

### DIFF
--- a/src/year2020/day06.rs
+++ b/src/year2020/day06.rs
@@ -6,39 +6,63 @@
 //! is only 26. For each yes answer we set a bit, shifting left based on the letter. For example
 //! `acf` would be represented as `100101`.
 //!
+//! The two lines that update the bitset ( bit = 1 << (b & 32); curr |= bit;) can be combined
+//! into a single opcode on x86, BTS which combines them into a single-cycle instruction.
+//!
 //! For part one to find groups where any person answered yes, we reduce the group using
 //! [bitwise OR](https://en.wikipedia.org/wiki/Bitwise_operation) then count the number of ones
-//! for each group using the blazing fast [`count_ones`] intrinsic.
+//! for each group using the blazing fast [`count_ones`] intrinsic. (popcount on x86)
 //!
 //! Part two is very similar, except that we use a bitwise AND instead.
 //!
-//! [`day 5`]: crate::year2020::day05
+//! [`day 6`]: crate::year2020::day06
 //! [`count_ones`]: u32::count_ones
 
-pub fn parse(input: &str) -> Vec<u32> {
-    input.lines().map(|line| line.bytes().fold(0, |acc, b| acc | (1 << (b - b'a')))).collect()
-}
+type Input = [u32;2];
 
-pub fn part1(input: &[u32]) -> u32 {
-    count_answers(input, u32::MIN, |group, passenger| group | passenger)
-}
+pub fn parse(input: &str) -> Input 
+{
+    let mut input = input.to_string();
+    if input.as_bytes()[input.len()-1] != b'\n' { input.push('\n')}
+    input.push('\n');
 
-pub fn part2(input: &[u32]) -> u32 {
-    count_answers(input, u32::MAX, |group, passenger| group & passenger)
-}
-
-fn count_answers(input: &[u32], initial: u32, combine: fn(u32, u32) -> u32) -> u32 {
-    let mut total = 0;
-    let mut group = initial;
-
-    for &passenger in input {
-        if passenger == 0 {
-            total += group.count_ones();
-            group = initial;
-        } else {
-            group = combine(group, passenger);
+    let bytes = input.as_bytes();
+    let mut i = 0;
+    let mut part1 = 0;
+    let mut part2 = 0;
+    let mut all:u32 = u32::MAX;
+    let mut any:u32 = 0;
+    let mut curr = 0;
+    let len = bytes.len();
+    loop {
+        let mut b = bytes[i]; i += 1;
+        loop {
+            let bit = 1 << (b & 31);
+            curr |= bit;
+            b = bytes[i]; i += 1;
+            if b == b'\n' {break}
+        }
+        all &= curr;
+        any |= curr;
+        curr = 0;
+        if bytes[i] == b'\n' { // Is this a double newline ending a section? Safe look-ahead due to input padding!
+            part1 += any.count_ones();
+            part2 += all.count_ones();
+            all = u32::MAX;
+            any = 0;
+            i += 1;
+            if i >= len {break} // The input can only end here, so the main loop does not have to test
         }
     }
+    [part1, part2]
 
-    total + group.count_ones()
 }
+
+pub fn part1(input: &Input) -> u32 {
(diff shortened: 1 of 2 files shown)


src/year2025/day05.rs
```diff
@@ -4,49 +4,59 @@
 //! complexity by first sorting the ranges in ascending order of their start.
 //!
```
maneatingape, line 5: Revert this file, e.g.

`git checkout main -- src/year2025/day05.rs`

or

`git restore --source=main src/year2025/day05.rs`

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
-    input.lines().map(|line| line.bytes().fold(0, |acc, b| acc | (1 << (b - b'a')))).collect()
-}
 
-pub fn part1(input: &[u32]) -> u32 {
-    count_answers(input, u32::MIN, |group, passenger| group | passenger)
+type Input = [u32;2];
+
+pub fn parse(input: &str) -> Input 
+{
+    let mut input = input.to_string();
```
maneatingape, line 25: We can avoid this mutable clone, by performing a final update to `part1` and `part2` after the main loop (see later comment)

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+    input.push('\n');
+
+    let bytes = input.as_bytes();
+    let mut i = 0;
+    let mut part1 = 0;
+    let mut part2 = 0;
+    let mut all:u32 = u32::MAX;
+    let mut any:u32 = 0;
+    let mut curr = 0;
+    let len = bytes.len();
```
maneatingape, line 36: Simplify manual loop with `for b in input.bytes() {`
removing `i` and `len`.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+    if input.as_bytes()[input.len()-1] != b'\n' { input.push('\n')}
+    input.push('\n');
+
+    let bytes = input.as_bytes();
+    let mut i = 0;
+    let mut part1 = 0;
+    let mut part2 = 0;
+    let mut all:u32 = u32::MAX;
+    let mut any:u32 = 0;
+    let mut curr = 0;
```
maneatingape, line 35: Minor nit: Rename `curr` => `mask` to align style with other days.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+{
+    let mut input = input.to_string();
+    if input.as_bytes()[input.len()-1] != b'\n' { input.push('\n')}
+    input.push('\n');
+
+    let bytes = input.as_bytes();
+    let mut i = 0;
+    let mut part1 = 0;
+    let mut part2 = 0;
+    let mut all:u32 = u32::MAX;
```
maneatingape, line 33: Minor nit: Remove explicit `u32` type, this can be inferred.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+    let mut any:u32 = 0;
+    let mut curr = 0;
+    let len = bytes.len();
+    loop {
+        let mut b = bytes[i]; i += 1;
+        loop {
+            let bit = 1 << (b & 31);
```
maneatingape, line 40: `1 << (b - b'a')` shows intent better.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+        if bytes[i] == b'\n' { // Is this a double newline ending a section? Safe look-ahead due to input padding!
+            part1 += any.count_ones();
+            part2 += all.count_ones();
+            all = u32::MAX;
+            any = 0;
+            i += 1;
+            if i >= len {break} // The input can only end here, so the main loop does not have to test
+        }
+    }
+    [part1, part2]
```
maneatingape, line 57: Insert

```
    part_one += any.count_ones();
    part_two += all.count_ones();
```

to avoid needing mutable string by treating end of string as a section break too.

This will break the test in `tests/year2020/day06.rs` so fix by just adding a trainling new line to `EXAMPLE`.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+
+pub fn parse(input: &str) -> Input 
+{
+    let mut input = input.to_string();
+    if input.as_bytes()[input.len()-1] != b'\n' { input.push('\n')}
+    input.push('\n');
+
+    let bytes = input.as_bytes();
+    let mut i = 0;
+    let mut part1 = 0;
```
maneatingape, line 31: Minor nit: Rename `part1` to `part_one`, `part2` to `part_two` to align style with other days.

src/year2020/day06.rs
```diff
@@ -6,26 +6,63 @@
...
+        loop {
+            let bit = 1 << (b & 31);
+            curr |= bit;
+            b = bytes[i]; i += 1;
+            if b == b'\n' {break}
+        }
+        all &= curr;
+        any |= curr;
+        curr = 0;
+        if bytes[i] == b'\n' { // Is this a double newline ending a section? Safe look-ahead due to input padding!
```
maneatingape, line 48: We can avoid the look ahead and simplify since `any` is 0 if this is a empty block, e.g.

```
if b == b`\n' {
    if any == 0 {
        // Ending a section
   } else {
       // Ending line within section
    }
} else {
    // Update mask 
}
```

<Optional> Rust also has an `match` (similar but more expressive than C's `switch`) that can express the same thing more succintly, e.g.

```
        match b {
            b'\n' if mask == 0 => {
                // Ending a section
            }
            b'\n' => {
                // Ending line within section
            }
            // Undercore is equivalent to `default:`
            _ => // Update mask
        }
```

maneatingape (changes requested):
Thanks for the PR! Already much faster than the original and we can still shave some more time off.

src/year2020/day06.rs
ebblake: Yes, but masking is faster than subtracting by a non power of 2, especially on x86 where the compiler can elide the mask because the cpu implicitly masks. That is, bits 1-26 are more efficient than 0-25.

ebblake:
Compare to my similar attempt in #83

src/year2020/day06.rs
ebblake: My alternative used `.chain(once(b'\n'))` to keep input immutable while avoiding the post-loop cleanup.

src/year2020/day06.rs
ebblake: I got more performance with a separate bool than by directly checking if any was 0

PR #85, opened 2026-06-14
2020 day12 performance 

TerjeWiigMathisen:
## Description

2020 / day12  8x speedup from a custom single-pass parser
Please ignore the spurious day06 commit, I don't know how to get rid of it!

## Type of change

- [x] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [x] Tests pass `cargo test`


src/year2020/day06.rs
```diff
@@ -28,4 +28,4 @@ pub fn part2(input: &[u32]) -> u32 {
 
 fn count_answers(input: &[u32], initial: u32, combine: fn(u32, &u32) -> u32) -> u32 {
     input.split(|&p| p == 0).map(|group| group.iter().fold(initial, combine).count_ones()).sum()
-}
+}
```
maneatingape: The day 6 PR had a nice speedup and I think we can get it merged with some cleanup, if you'd like to re-open that PR.

src/year2020/day12.rs
```diff
@@ -1,62 +1,92 @@
...
-            b'S' => position.y += amount,
-            b'E' => position.x += amount,
-            b'W' => position.x -= amount,
-            b'L' => direction = rotate(direction, -amount),
-            b'R' => direction = rotate(direction, amount),
-            b'F' => position += direction * amount,
-            _ => unreachable!(),
-        }
+pub fn parse(input: &str) -> Input {
+    let mut inp = input.to_string();
```
maneatingape: `parse.rs` does exactly this byte by byte parsing, exposed as 2 neat typed iterators (one for unsigned number and one for signed)  so that you can write:

```
use crate::util::parse::*;

let amounts = input.input.signed();
```

Since we also need the direction command use the nifty methods on `Iterator` to compose 2 iterators into a single iterator over a 2-tuple, e.g.

```
    let first = input.bytes().filter(u8::is_ascii_uppercase);
    let second = input.iter_signed();
```

We can then use this (without an intermediate `vec`) in your core loop that combines part one and part two, e.g. something like:

```
    let mut part_one = ORIGIN;
    let mut part_two = ORIGIN;
    let mut direction = RIGHT;
    let mut waypoint = Point::new(10, -1);

    for (command, amount) in first.zip(second) {
        match command {
            b'N' => {
                part_one.y -= amount;
                waypoint.y -= amount;
            }
           ...
```

src/year2020/day12.rs
```diff
@@ -1,62 +1,92 @@
...
+    (part1, part2)
+}
 
-    position.manhattan(ORIGIN)
+pub fn part1(input: &Input) -> i32 {
+    input.0
 }
 
-fn rotate(point: Point, amount: i32) -> Point {
-    match amount.rem_euclid(360) {
```
maneatingape, line 56: Can omit the `rem_euclid` if the class from `L` reflect the rotation e.g. `direction = rotate(direction, 360 - amount);`

maneatingape (changes requested):
Nice, I get a roughly 3x speedup.

We can re-use existing parsing utility methods to get the same effect with less code, while keeping the core idea.

src/year2020/day06.rs
ebblake: Or we can use #83 for the day 6 speedup.

maneatingape:
This was a nice performance increase, so I implemented the suggestions then merged the PR on your behalf.

PR #87, opened 2026-06-20
Handle newlines more efficiently than input.lines()

ebblake:
## Description

By tracking our own stacks instead of using a recursion call stack, it is possible to solve part 1 and 2 in parallel as part of the parse. Provides a slight speedup; on my machine, I measured 82us pre-patch, and 72us with the patch included.  The testsuite needs a trailing newline to match expectations.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day18.rs
+++ b/src/year2020/day18.rs
@@ -11,14 +11,75 @@
 //! * 1 + 2 * 3 + 4 => 1 + 2 * (3 + 4)
 //! * 1 + (2 * 3 * 4) + 5 => 1 + (2 * (3 * (4))) + 5
 use crate::util::parse::*;
-use std::str::Bytes;
+use std::slice::Iter;
 
-pub fn parse(input: &str) -> Vec<&str> {
-    input.lines().collect()
+/// A custom iterator treats each line of input as a parenthesized expression to add, so that
+/// the solver operates over the entire buffer without breaking it into lines.
+struct FlatIter<'a> {
+    bytes: Iter<'a, u8>,
+    state: State,
 }
 
-pub fn part1(input: &[&str]) -> u64 {
-    fn helper(bytes: &mut Bytes<'_>) -> u64 {
+#[derive(Clone, Copy)]
+enum State {
+    Start, // Need opening '(' for start of line.
+    Body,  // Emitting normal bytes from within a line.
+    End,   // Just supplied closing ')' at end of line.
+    Plus,  // Emitting '+' between lines.
+}
+
+impl<'a> FlatIter<'a> {
+    fn new(input: &'a [u8]) -> Self {
+        Self { bytes: input.iter(), state: State::Start }
+    }
+}
+
+impl Iterator for FlatIter<'_> {
+    type Item = u8;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.state {
+            State::Start => {
+                if self.bytes.len() == 0 {
+                    return None;
+                }
+                self.state = State::Body;
+                Some(b'(')
+            }
+            State::Body => {
+                match self.bytes.next() {
+                    // The expressions are consistently formatted so encountering a space just means
+                    // skip and return the next character that will always be present.
+                    Some(&b' ') => self.bytes.next().copied(),
+                    Some(&b'\n') | None => {
+                        self.state = State::End;
+                        Some(b')')
+                    }
+                    other => other.copied(),
+                }
+            }
+            State::End => {
+                if self.bytes.len() == 0 {
+                    None
+                } else {
+                    self.state = State::Plus;
+                    Some(b'+')
+                }
+            }
+            State::Plus => {
+                self.state = State::Body;
+                Some(b'(')
+            }
+        }
+    }
+}
+
+pub fn parse(input: &str) -> &str {
+    input
+}
+
+pub fn part1(input: &str) -> u64 {
+    fn helper(bytes: &mut FlatIter<'_>) -> u64 {
         let mut total = value(bytes, helper);
 
         while let Some(operation) = next(bytes) {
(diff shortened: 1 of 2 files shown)


maneatingape:
The original solution is very short and understandable and the implicit stack via recursion approach is very slightly faster than this dual stack.

Where it falls down is the very slow initial parsing `input.lines().collect()`. This takes up 20% of the total time.

I feel we could preserve much of the original simplicity but add special handling for each `\n` character so that we could treat the input as a single stream of bytes, elimating the slow up-front parsing.

ebblake:
> The original solution is very short and understandable and the implicit stack via recursion approach is very slightly faster than this dual stack.
> 
> Where it falls down is the very slow initial parsing `input.lines().collect()`. This takes up 20% of the total time.
> 
> I feel we could preserve much of the original simplicity but add special handling for each `\n` character so that we could treat the input as a single stream of bytes, elimating the slow up-front parsing.

Fair enough - I'll give that approach a try

ebblake:
> > The original solution is very short and understandable and the implicit stack via recursion approach is very slightly faster than this dual stack.
> > Where it falls down is the very slow initial parsing `input.lines().collect()`. This takes up 20% of the total time.
> > I feel we could preserve much of the original simplicity but add special handling for each `\n` character so that we could treat the input as a single stream of bytes, elimating the slow up-front parsing.
> 
> Fair enough - I'll give that approach a try

Here's my attempt - but it is running slower than pre-patch, so I'm probably not doing it quite the way you envisioned:

```
diff --git i/src/year2020/day18.rs w/src/year2020/day18.rs
index 01a6fa9..33a15e3 100644
--- i/src/year2020/day18.rs
+++ w/src/year2020/day18.rs
@@ -12,16 +12,21 @@
 //! * 1 + (2 * 3 * 4) + 5 => 1 + (2 * (3 * (4))) + 5
 use crate::util::parse::*;
 use std::str::Bytes;
+use std::iter::Peekable;

-pub fn parse(input: &str) -> Vec<&str> {
-    input.lines().collect()
+pub fn parse(input: &str) -> &str {
+    input
 }

-pub fn part1(input: &[&str]) -> u64 {
-    fn helper(bytes: &mut Bytes<'_>) -> u64 {
+pub fn part1(input: &str) -> u64 {
+    fn helper(bytes: &mut Peekable<Bytes<'_>>, nested: bool) -> u64 {
         let mut total = value(bytes, helper);

-        while let Some(operation) = next(bytes) {
+        while let Some(operation) = next(bytes, nested) {
+            if operation == b'\n' {
+                total += helper(bytes, false);
+                break;
+            }
             let value = value(bytes, helper);
             if operation == b'+' {
                 total += value;
@@ -33,34 +38,42 @@ pub fn part1(input: &[&str]) -> u64 {
         total
     }

-    input.iter().map(|line| helper(&mut line.bytes())).sum()
+    helper(&mut input.bytes().peekable(), false)
 }

-pub fn part2(input: &[&str]) -> u64 {
-    fn helper(bytes: &mut Bytes<'_>) -> u64 {
+pub fn part2(input: &str) -> u64 {
+    fn helper(bytes: &mut Peekable<Bytes<'_>>, nested: bool) -> u64 {
         let mut total = value(bytes, helper);

-        while let Some(operation) = next(bytes) {
+        while let Some(operation) = next(bytes, nested) {
             if operation == b'+' {
                 total += value(bytes, helper);
+            } else if operation == b'\n' {
+                total += helper(bytes, false);
+                break;
             } else {
                 // Implicitly insert '(' and ')' around the remaining sub-expression so when it
                 // finishes we break too.
-                total *= helper(bytes);
-                break;
+                total *= helper(bytes, true);
+                if nested {
+                    break;
+                }
             }
         }

         total
     }

-    input.iter().map(|line| helper(&mut line.bytes())).sum()
+    helper(&mut input.bytes().peekable(), false)
 }

 /// Convenience wrapper around [`Bytes`] iterator. Encountering a `)` is also considered end of
 /// sequence. The expressions are consistently formatted so encountering a space just means
 /// skip and return the next character that will always be present.
-fn next(bytes: &mut Bytes<'_>) -> Option<u8> {
+fn next(bytes: &mut Peekable<Bytes<'_>>, nested: bool) -> Option<u8> {
+    if nested && bytes.peek() == Some(&b'\n') {
+        return None;
+    }
     match bytes.next() {
         None | Some(b')') => None,
         Some(b' ') => bytes.next(),
@@ -70,9 +83,10 @@ fn next(bytes: &mut Bytes<'_>) -> Option<u8> {

 /// Convenience wrapper to return the value of either the next raw digit literal or a
 /// sub-expression nested in parentheses.
-fn value(bytes: &mut Bytes<'_>, helper: fn(&mut Bytes<'_>) -> u64) -> u64 {
-    match next(bytes).unwrap() {
-        b'(' => helper(bytes),
-        b => b.to_decimal() as u64,
+fn value(bytes: &mut Peekable<Bytes<'_>>, helper: fn(&mut Peekable<Bytes<'_>>, bool) -> u64) -> u64 {
+    match next(bytes, true) {
+        None => 0,
+        Some(b'(') => helper(bytes, true),
+        Some(b) => b.to_decimal() as u64,
     }
 }
```

If there is some easy way to filter the main iterator so that we wrap it in `(` at front, `0)` at end, and then convert all `\n` in the middle to `)+(`, then the existing code would handle that munged iterator without any special casing of \n.

ebblake:
> If there is some easy way to filter the main iterator so that we wrap it in `(` at front, `0)` at end, and then convert all `\n` in the middle to `)+(`, then the existing code would handle that munged iterator without any special casing of \n.

With some help from AI, I created a custom iterator that wraps an underlying .to_bytes() iterator, and that one had a speedup. Revising the patch along those lines.

maneatingape:
Tried a few different approaches. Your initial dual stack approach was ~20% faster by eliminating the `lines()` splitting, but more complex. Adding a tertiary newline check to the recursive approach did work, but then messed up branch prediction so that it was overall slower than splitting each line first.

On balance I'm going to stick with the simple recursive approach for this solution, given the speed/complexity tradeoff.

PR #88, opened 2026-06-25
Use Pohlig-Hellman computation for a 10x speedup.

ebblake:
## Description

Instead of computing 4495 table entries for baby-step-giant-step over the full √20201227 space, we can further exploit Pohlig-Hellman to divide the problem across the prime divisors of 20201226.  For 2, 3, and 29, the computations are small enough to brute force and do table lookups without any thought; for 116099, baby-step-giant-step is still worthwhile, but we can now stop at √116099 = 340 iterations.  Plus, this smaller size is actually worth computing at compile-time. Combine the four answers using Chinese Remainder Theorem.

On my laptop, this speeds up compuation from 31.3us to 2.7us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [x] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day25.rs
+++ b/src/year2020/day25.rs
@@ -1,17 +1,45 @@
 //! # Combo Breaker
 //!
-//! The card loop size is found using the
-//! [Baby-step giant-step algorithm](https://en.wikipedia.org/wiki/Baby-step_giant-step).
-//! This takes only √20201227 = 4495 steps, compared to potentially up to 20201227 steps
-//! for the brute force approach.
+//! The puzzle description for today describes the [Diffie-Hellman-Merkle] handshake, which your
+//! browser uses every day for encrypted traffic. It relies on the fact that if you use the
+//! product of two prime numbers as your encryption key, where each of the two parties in
+//! the transaction only knows one of the two prime numbers, then modular arithmetic allows
+//! both sides to efficiently compute the same results, while an eavesdropper would have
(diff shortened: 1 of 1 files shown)


ebblake:
I blame askalski for teaching me this, although his solution goes even crazier with SIMD: https://github.com/Voltara/advent2020-fast/blob/main/src/day25.cpp

maneatingape (commented):
Never heard about the Pohlig-Hellman algorithm before, so this was a nice learning.

I feel that this refactor is too complex for the speed up, given the original solution is reasonably quick and very understandable.

Happy to add a paragraph comment mentioning Pohlig-Hellman, and additionally link to either this PR or to your own repo, if you'd like.

ebblake:
> Never heard about the Pohlig-Hellman algorithm before, so this was a nice learning.
> 
> I feel that this refactor is too complex for the speed up, given the original solution is reasonably quick and very understandable.

Makes sense, and it's your repo so I'll live with your decision. The jump from O(n) to O(sqrt n) is much easier to code/grasp than Pohlig-Hellman that merely reduces the size of n.

> Happy to add a paragraph comment mentioning Pohlig-Hellman, and additionally link to either this PR or to your own repo, if you'd like.

I'll spin up a second PR with just a docs patch, then close this one, so the link in that one can still show this as the faster but more complicated patch.

PR #89, opened 2026-06-26
Short-circuit the search by exploiting the sorted array

ebblake:
## Description

A recent patch sorted the array of numbers on the grounds that the average runtime improves when visiting pairs in ascending order.  But we can go one further: the mere act of creating our hash for the part one answer is a radix sort.  Even though walking O(buckets) after an O(n) radix sort (2020 buckets if we iterate them all) is slightly slower than an O(n log n) item sort (with n=200, this is about 200*8=1600 comparisons), we can additionally take advantage of the fact that things are sorted by exploiting the triangle inequality: at least one of the items will be at least 2020/3, and another will be less than the integer we are using as our pivot point.  With a simple enough loop, the compiler can see the loop invariants and hoist 0 checks earlier; and even though the loop is still quadratic, input files have fewer low entries so this completes even faster.  With this, parse() speeds up because it is no longer doing a duplicate O(n log n) sort, part1() speeds up because it is now done as a side effect of parse, and part2() is slightly faster.
    
On my laptop, this changes parse/part1/part2 runtime from 2.0/0.3/0.4 (2.7us total) to 0.8/0.0/0.2 (1.0us total), less than half the time.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [x] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day01.rs
+++ b/src/year2020/day01.rs
@@ -8,50 +8,49 @@
 //! For part one we use an implicit hash table in an array, since values are constrained to between
 //! 0 and 2020 and each value is already perfectly hashed. For each entry we check the index
 //! at its value. If this is marked then we have seen the reciprocal `2020 - value` before
-//! so we can return the product. If not then we mark the reciprocal value in the array.
+//! so we have found the answer. Creating this array also performs a radix sort that will
+//! be used in part two.
 //!
-//! Part two reuses the pair finding logic, finding the third element by stepping through the slice
-//! one by one and adjusting the target total. To reuse the array without reallocating
-//! (which is slow) we use a `round` value instead of `bool`.
+//! Part two adds a second array, this time containing the product of any two numbers that
+//! sum to that point. Since the problem only has one correct answer, it does not matter if
+//! other array slots get written more than once. Because the list is already sorted, we can short
+//! circuit iterations as soon as a sum is not possible.
 use crate::util::parse::*;
 
-pub fn parse(input: &str) -> Vec<usize> {
-    let mut numbers: Vec<usize> = input.iter_unsigned().collect();
-    // Assume that, after sorting, the solutions are closer to the middle of the vector than the
-    // extremes. It's not necessary for correctness but improves the average running time for
-    // most (all?) inputs.
-    numbers.sort_unstable();
-    numbers
+type Input = (usize, [bool; 2020]);
+
+pub fn parse(input: &str) -> Input {
+    let mut numbers = [false; 2020];
+    let mut part_one = 0;
+
+    // Part one is determined as a side effect of the parse. Assume the input has no duplicates.
+    for number in input.iter_unsigned::<usize>() {
+        if numbers[2020 - number] {
+            part_one = number * (2020 - number);
+        }
+        numbers[number] = true;
+    }
+
+    // The parse performed a radix sort; numbers can now be used as an ordered sparse array.
+    (part_one, numbers)
 }
 
-pub fn part1(input: &[usize]) -> usize {
-    let mut hash = [0; 2020];
-    two_sum(input, 2020, &mut hash, 1).unwrap()
+pub fn part1(input: &Input) -> usize {
+    input.0
 }
 
-pub fn part2(input: &[usize]) -> usize {
-    let mut hash = [0; 2020];
+pub fn part2(input: &Input) -> usize {
+    let (_, numbers) = input;
 
-    (0..input.len() - 2)
-        .find_map(|i| {
-            let first = input[i];
-            let round = i + 1;
-            let target = 2020 - first;
-            two_sum(&input[round..], target, &mut hash, round).map(|product| first * product)
-        })
-        .unwrap()
-}
-
-fn two_sum(slice: &[usize], target: usize, hash: &mut [usize], round: usize) -> Option<usize> {
-    for &i in slice {
-        if i < target {
-            let complement = target - i;
-            if hash[i] == round {
-                return Some(i * complement);
+    // We know at least one of the three numbers is at least 2020/3.
+    for i in 674..2020 {
+        for j in 1..i {
+            let k = 2020 - i - j;
+            if numbers[i] && numbers[j] && numbers[k] {
+                return i * j * k;
             }
-            hash[complement] = round;
         }
     }
 
-    None
+    unreachable!()
 }


src/year2020/day01.rs
```diff
@@ -8,50 +8,63 @@
...
-}
-
-fn two_sum(slice: &[usize], target: usize, hash: &mut [usize], round: usize) -> Option<usize> {
-    for &i in slice {
-        if i < target {
-            let complement = target - i;
-            if hash[i] == round {
-                return Some(i * complement);
+    // The input only had 3- and 4-digit values.
+    for i in 100..2020 {
```
ebblake: Since we guarantee an early return, maybe it's better to write this as `let mut i = 100; loop {`

src/year2020/day01.rs
```diff
@@ -8,50 +8,63 @@
...
+        if numbers[i] {
+            // Short-circuit the inner loop as soon as a sum is impossible.
+            let bound = (2020 - i) / 2;
+            for j in i + 1..bound {
+                if numbers[j] {
```
ebblake: This spends time rechecking for non-empty elements of numbers. I wonder if a pre-pass to consolidate from a sparse array back into a dense list would help.

src/year2020/day01.rs
```diff
@@ -8,50 +8,63 @@
...
+                    // Check for either i matching to a previous pair, or the pair i and j
+                    // matching to another input.
+                    if hash[2020 - i] != 0 {
```
ebblake: This check does not depend on j; hoist it out of the inner loop.

src/year2020/day01.rs
ebblake: I tried:
```
+    let mut numbers = Vec::with_capacity(200);
+    numbers.extend(hash.iter().enumerate().skip(100).filter(|&(_, &present)| present).map(|(idx, _)| idx));
```
but using numbers proved slower than just using hash.

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
-        })
-        .unwrap()
-}
+    // The input only had 3- and 4-digit values. Given i < j < k, the outer loop iterates
+    // over j. Since an answer exists, we don't need bounds on j, but if we wanted, 100..960
+    // is appropriate, since j can never be more than 2020-k-i. Although the search for j covers
+    // a sparse array, we can build up the array of valid i to make the inner loop faster.
+    let mut seen = [0u16; 200];
+    let mut seen_ptr = 0;
+    let mut j = 100;
```
ebblake: I found a different input file with low point of 71. Starting my read of the buckets at 100 worked on my input but is not reliable; I'll rework to start from 0.

src/year2020/day01.rs
ebblake: Flattening the list up front was a bust, but creating a flat list for i while iterating over j on the outer loop was a win.

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
-    // most (all?) inputs.
-    numbers.sort_unstable();
-    numbers
+type Input = (usize, [bool; 2020]);
+
+pub fn parse(input: &str) -> Input {
+    let mut numbers = [false; 2020];
+    let mut part_one = 0;
+
+    // Part one can be determined as a side effect of the parse. Assume only one correct answer.
```
maneatingape: > Assume only one correct answer.

If there was more than one combination the puzzle text would provide a way to tie break, so I don't think need to mention this.

What I would mention instead in the comment, is that this assumes that each number is unique, e.g there is never a valid solution such as 10 * 10 * 2000

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
-        .unwrap()
-}
+    // Given i < j < k, the outer loop iterates over j. Since an answer exists, we don't need
+    // bounds on j, but if we wanted, 0..1010 is appropriate, since j can never be more than
+    // 2020-k-i. Although the search for j covers a sparse array, we can build up the array of
+    // valid i to make the inner loop faster.
+    let mut seen = [0_u16; 200];
+    let mut seen_ptr = 0;
+    let mut j = 0;
+    loop {
```
maneatingape: Replace manual for loop with `for j in 0..1010`

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
+                // Check for the pair i and j matching to another input.
+                let k = 2020 - i - j;
+                if numbers[k] {
+                    return i * j * k;
+                }
+                // Store the partial product as a witness of a pair.
+                hash[i + j] = i * j;
             }
-            hash[complement] = round;
+            seen[seen_ptr] = j as u16;
```
maneatingape: Replace manual vec implementation with:
```
let mut seen = Vec::with_capacity(200);
...
seen.push(j)
```

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
+            }
 
-fn two_sum(slice: &[usize], target: usize, hash: &mut [usize], round: usize) -> Option<usize> {
-    for &i in slice {
-        if i < target {
-            let complement = target - i;
-            if hash[i] == round {
-                return Some(i * complement);
+            // Short-circuit the inner loop as soon as a sum is impossible.
+            let bound = (2020 - j) >> 1;
```
maneatingape: ` let bound = (2020 - j) / 2;` expresses the intent better. The compiler will lower this to the bitshift.

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
+            for i in seen.iter().take(seen_ptr) {
```
maneatingape: `for i in seen.iter().take_while(|&&i| i < bound)`

src/year2020/day01.rs
```diff
@@ -8,50 +8,75 @@
...
+    // bounds on j, but if we wanted, 0..1010 is appropriate, since j can never be more than
+    // 2020-k-i. Although the search for j covers a sparse array, we can build up the array of
+    // valid i to make the inner loop faster.
+    let mut seen = [0_u16; 200];
+    let mut seen_ptr = 0;
+    let mut j = 0;
+    loop {
+        if numbers[j] {
+            // Check for j matching to a previous pair.
+            if hash[2020 - j] != 0 {
```
maneatingape: Removing the hash code *sped up* my input by 200ns, does it have the same impact on yours?

maneatingape (changes requested):
Nice trick to shave off even more time.

maneatingape:
Taking the core idea of using sparse array, we can simplify part two to just:

```
pub fn part2(input: &Input) -> usize {
    let (_, numbers) = input;

    for i in 674..2020 {
        for j in 1..i {
            let k = 2020 - i - j;
            if numbers[i] && numbers[j] && numbers[k] {
                return i * j * k;
            }
        }
    }

    unreachable!()
}
```

(the compiler hoists the `numbers[i]` check out of the inner loop)

Performance is 300ns vs 400ns for the patch on my machine.

src/year2020/day01.rs
ebblake: That's interesting - I started with a Vec and swapped to a hard-coded array to see if it would avoid overhead of memory allocation. If it doesn't penalize performance, I will go with the vec.

ebblake:
> 674..2020

How did you pick 674 as your starting point? 2020/3?

maneatingape:
> > 674..2020
> 
> How did you pick 674 as your starting point? 2020/3?

Yup, my logic is that at least one of the 3 numbers has to be > 2020 / 3

src/year2020/day01.rs
ebblake: or ditch it altogether with your approach which was even faster ;)

src/year2020/day01.rs
ebblake: Yes. Nice find!

maneatingape (approved):
👍

PR #90, opened 2026-06-29
Document possible speed-up from Pohlig-Hellman

ebblake:
## Description

Since this solution is already under 30us, the complexity required to implement Pohlig-Hellman to further bring it down to 3us outweighs the maintainability.  However, it is worth documenting the possibility, and linking to a patch with the work available.

Replacement for #88 that just focuses on documenting the speedup.

## Type of change

- [ ] Performance improvement
- [ ] Bug fix
- [X] Other - documentation

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2020/day25.rs
+++ b/src/year2020/day25.rs
@@ -1,17 +1,44 @@
 //! # Combo Breaker
 //!
-//! The card loop size is found using the
-//! [Baby-step giant-step algorithm](https://en.wikipedia.org/wiki/Baby-step_giant-step).
-//! This takes only √20201227 = 4495 steps, compared to potentially up to 20201227 steps
-//! for the brute force approach.
+//! The puzzle description for today describes the [Diffie-Hellman-Merkle] handshake, which your
+//! browser uses every day for encrypted traffic. It relies on the fact that if you use the
+//! product of two prime numbers as your encryption key, where each of the two parties in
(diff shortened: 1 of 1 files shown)


ebblake:
Failing lint looks like a recent change in clippy rather than something this patch triggered.

ebblake:
Is it worth trying to salvage computing the baby-step table at compile-time from #88?  In the Pohlig-Hellman solution, the baby-step table only needs 341 slots (totally doable at compile time). Here, we're looking at 4495 entries, which is quite a chunk of memory (over 16k) and compile time just to allow an O(log n) search of the compiled table rather than a dynamic FastMap search (which is nominally O(1) search but has higher overhead due to setting up the FastMap at runtime).

maneatingape:
> Is it worth trying to salvage computing the baby-step table at compile-time from #88? In the Pohlig-Hellman solution, the baby-step table only needs 341 slots (totally doable at compile time). Here, we're looking at 4495 entries, which is quite a chunk of memory (over 16k) and compile time just to allow an O(log n) search of the compiled table rather than a dynamic FastMap search (which is nominally O(1) search but has higher overhead due to setting up the FastMap at runtime).

With 262 (and counting!) solutions, compile times are already stretching...I think your're right that the cost outweighs the benefit in this case.

maneatingape (approved):
👍

PR #91, opened 2026-06-30
Use frequency counters rather than repeated list iteration

ebblake:
## Description

Instead of parsing the input file into a list of integers, then iterating over that list once per each bit lane, we can use the parse to populate frequency counters for every bit lane. We can also build up a balanced binary tree backed by array storage, where every number seen populates a leaf node, and all parent nodes are the sum of their two children, with just one linear pass during parse.  From there, part one just reads the frequency bins (12 lookups) and part two is just a binary search (12 comparisons per rating).
    
On my laptop, this improves runtime from 35us down to under 12us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day03.rs
+++ b/src/year2021/day03.rs
@@ -1,22 +1,64 @@
 //! # Binary Diagnostic
 //!
-//! Part one uses bit manipulation to build up the binary numbers directly one digit at a time.
+//! Part one collects 12 frequency counters, one for each digit lane. Each line of input increments
+//! the corresponding counters for each `1`. The number of `0` in the lane can be recovered by
+//! subtraction, making it easy to construct both the gamma and epsilon rate.
 //!
-//! Part two clones the input `vec` then uses [`retain`] to efficiently discard numbers that
-//! don't meet the criteria.
-//!
-//! [`retain`]: Vec::retain
-pub fn parse(input: &str) -> Vec<&[u8]> {
-    input.lines().map(str::as_bytes).collect()
+//! Part two constructs a binary tree, stored as an array. Using an array of 2*4096 bins is enough
+//! to treat each of the 1000 unique inputs as a leaf in the second half of the array, and then
+//! each node in the first half is computed as the sum of its two children from the right half. The
+//! array can be built in linear time, at which point finding the oxygen generator rating and
+//! CO2 scrubber rating are each a binary search in the tree.
+pub struct Input {
+    entries: u16,    // Number of entries in the file.
+    lanes: Vec<u16>, // Contains size counters, one for each digit lane.
+    tree: Vec<u16>,  // Contains a balanced binary tree with 2**(width+1) nodes.
 }
 
-pub fn part1(input: &[&[u8]]) -> u32 {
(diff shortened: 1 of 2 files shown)


maneatingape:
I like the insight. Part two is essentially asking us to walk a binary tree.  Another input where the unspecified constraint is that each number is unique.

We can go even faster (~20%) and simpler by building a tree starting with the leaves, 1 if a number if present, 0 if not.
Then each level up is the sum of its children.

The for part two, walk the tree from the root + 1 level and the number you arrive at is the answer.

For example (excuse the hacky code):
```
    let size = input.lines().next().unwrap().len();

    let mut foo = vec![0; size];
    let mut bar: Vec<usize> = vec![0; 1 << (size + 1)];
    let mut offset = 1 << size;

    for chunk in input.as_bytes().chunks(size + 1) {
        let mut binary = 0;

        for (i, b) in chunk[..size].iter().enumerate() {
            let bit = (b & 1) as usize;
            foo[i] += bit;
            binary = (binary << 1) | bit;
        }

        bar[offset + binary] = 1;
    }

    while offset > 2 {
        let half = offset / 2;

        for i in half..offset {
            bar[i] = bar[2 * i] + bar[2 * i + 1];
        }

        offset = half;
    }

    let mut ogr = 2;

    let rows = input.len() / (size + 1);
    let mut gamma = 0;
    let mut epsilon = 0;
    
    for ones in foo {
        let zeros = rows - ones;
        gamma = (gamma << 1) | u32::from(ones > zeros);
        epsilon = (epsilon << 1) | u32::from(zeros > ones);
    }

    for _ in 0..size {
        ogr = 2 * ogr + if bar[ogr + 1] >= bar[ogr] { 2 } else { 0 };
    }

    let mut csr = 2;

    for _ in 0..size {
        csr = 2 * csr + if bar[csr + 1].wrapping_sub(1) < bar[csr].wrapping_sub(1) { 2 } else { 0 };
    }
    
    println!("Part one {} Part two {}", gamma * epsilon, (ogr / 2 - (1 << size)) * (csr / 2 - (1 << size)));
```

ebblake:
> I like the insight. Part two is essentially asking us to walk a binary tree. Another input where the unspecified constraint is that each number is unique.
> 
> We can go even faster (~20%) and simpler by building a tree starting with the leaves, 1 if a number if present, 0 if not. Then each level up is the sum of its children.

Nice! My prefix layout is also a form of binary tree (since we both used 8192 elements), but I'm performing 12 increments per line (12000) to populate it, while your variation only performs 1000 initial sets, followed by 4096 sums to build parents from children.  Both O(n), but yours has the better coefficient.  And both have the property of being able to find the right child node in 12 or fewer walks.  I'm happy to rework my code to pick up the ideas in your patch.

ebblake:
I didn't quite see the 20% gain you saw, but it was more than 10% (from 13us down to 11.5us). Patch updated.

maneatingape (approved):
👍

PR #92, opened 2026-07-01
Compute median with less effort

ebblake:
## Description

Instead of performing an O(n log n) sort_unstable, we can use the O(n) select_nth_unstable to get at the median element. Since the input size is always even (1000 crabs in the actual input, 10 in the example), we need the average between two elements - but it is also O(n) to grab the neighbor once we have the array segregated by the median element.

On my laptop, this cuts part1() from 6.6us to 2.4us (overall day drops from 11us to 7us).

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day07.rs
+++ b/src/year2021/day07.rs
@@ -29,10 +29,13 @@ pub fn part2(input: &[i32]) -> i32 {
 
 fn median(input: &[i32]) -> i32 {
     let mut crabs = input.to_vec();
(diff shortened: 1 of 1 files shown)


ebblake:
Similar to #67

maneatingape (approved):
👍

PR #93, opened 2026-07-04
Another median where quickselect is faster

ebblake:
## Description

O(n) select_nth_unstable is faster than O(n log n) sort_unstable.

On my laptop, this speeds up from 42us to 36us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`


maneatingape:
Ha, great minds think alike...I just caught this in a general clean up pass in year 2021.

Closing because the change has already been applied.

PR #94, opened 2026-07-06
Use modular math to skip intermediate folds

ebblake:
## Description

All input files for day 13 have the same fold pattern of several cuts in half, because they all generate the same size output buffer for letters.  Rather than performing each fold in succession, we can observe that mapping of final points to their original locations is periodic.  For example, along the y axis we know that this 14-byte pattern is repeated across all larger folds:
 abcdef_fedcba_
Thus, we can skip directly from the input point to the final fold by using the right period along that dimension before a final collapse into the desired end position.

On my laptop, this sped up part2() from 55us to 6us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day13.rs
+++ b/src/year2021/day13.rs
@@ -7,8 +7,10 @@
 //! We will use both approaches for speed, the first in part one and the second in part two.
 //!
 //! For part two we can determine the final size of the paper by taking the *last* x and y
-//! coordinates from the fold instructions. It's then faster and more convenient to process
-//! each point completely and update the final location, than to step through intermediate folds.
+//! coordinates from the fold instructions. Because each fold cuts the remaining paper in
+//! half, the layout of points that map to the same condensed positions is periodic. It's faster
+//! and more convenient to process each point completely using modular math to go straight to
+//! the condensed position without worrying about any intermediate folds.
 use crate::util::grid::*;
 use crate::util::hash::*;
 use crate::util::iter::*;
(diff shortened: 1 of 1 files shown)


src/year2021/day13.rs
```diff
@@ -9,6 +9,8 @@
 //! For part two we can determine the final size of the paper by taking the *last* x and y
 //! coordinates from the fold instructions. It's then faster and more convenient to process
 //! each point completely and update the final location, than to step through intermediate folds.
+//! Since each fold cuts the remaining paper in half, we can use modular math to bypass the
```
maneatingape: ❤️

src/year2021/day13.rs
```diff
@@ -64,7 +66,17 @@ pub fn part2(input: &Input) -> String {
     let mut grid = Grid::new(width + 1, height, '.');
 
     for &start in &input.points {
-        let end = input.folds.iter().fold(start, |point, &fold| apply_fold(fold, point));
+        let period_x = 2 * (width + 1);
```
maneatingape: Extract `period_x` and `period_y` out of the loop, since they don't change.

(The compiler will do this anyway but better to be explicit)

src/year2021/day13.rs
```diff
@@ -64,7 +66,17 @@ pub fn part2(input: &Input) -> String {
...
+        let mut x = start.x % period_x;
```
maneatingape: Minor suggestion: We can use rebinding to keep things immutable, e.g.

```
let x = start.x % period.x;
let x = if x > width { period_x - x - 2 } else {x };
```

(same for y)

maneatingape (changes requested):
Nice insight!

maneatingape (approved):
👍

PR #95, opened 2026-07-06
Pre-compute impact factors at compile time

ebblake:
## Description

This puzzle is already blindingly fast. But we can make it faster with an 18k table of compile-time constants, by pre-computing a single fish's impact over N generations, so that the runtime only has to compute the dot product of the initial numbers of fish with the constants for the target day.

On my laptop, this cuts the runtime from 750ns to 310ns.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day06.rs
+++ b/src/year2021/day06.rs
@@ -6,26 +6,50 @@
 //! Another optimization trick is rather than modifying the array by removing the fish at day 0,
 //! then shifting each fish total down by 1, we can simply increment what we consider the
 //! head of the array modulo 9 to achieve the same effect in place.
-use crate::util::parse::*;
+//!
+//! What's more, the computation of how many fish result after N generations for a single starting fish
+//! can be determined up front.  With a table containing only `9*257` u64 entries (about 18k
+//! memory) computed at compile time, we can determine the two answers in constant runtime.
 
 type Input = [u64; 9];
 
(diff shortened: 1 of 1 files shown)


ebblake:
I fully suspect that you have no reason to apply this one - getting a 50% improvement over 1us still shows as 1us in your totals.  Thus, the increase in compile time outweighs the benefits.  But it was sure fun to write!

src/year2021/day06.rs
```diff
@@ -6,26 +6,49 @@
...
+        table[day + 1][0] += table[day][6];
+        day += 1;
+    }
+
+    table
+};
+
 pub fn parse(input: &str) -> Input {
     let mut fish = [0_u64; 9];
     input.iter_unsigned().for_each(|i: usize| fish[i] += 1);
```
ebblake: This parse can be made faster. Since every fish is one byte, separated by commas, you could just iter().step_by(2) and use masks to determine the index, rather than the overhead of iter_unsigned().

maneatingape:
This is a zany completely over-engineered solution — very much in the Advent of Code tradition! 😄 

That said I'm going to stick with the current minimal solution which is already < 1µs.

PR #96, opened 2026-07-07
Yet another faster computation of median

ebblake:
## Description

I experimented with a radix sort instead of quickselect. Since the input file has more crabs below 1000 than above (proof: exactly 500 3-digit and 500 4-digit crabs would be a file size of 4500, but input files are closer to 3900 bytes), a radix sort can find the median in under 2000 operations (1000 writes to a 2048-entry hash table, then < 1000 buckets crawled).  But it takes more code, so sticking with quickselect (whose best case is 2*n operations if the pivots are chosen well, so just over 2000 operations) is more maintainable with the difference in effort in the noise.  But more importantly, even though we have an even number of crabs (and thus two choices on either side of the true median), we know that any integer between those two boundaries (if the two crabs are not at the same position) will produce the same fuel score, so there is no need to compute the TRUE median; less work makes for a faster part 1.

And for part 2, we can prove that the correct answer will come from either truncating or rounding the mean to the nearest integer; it is not necessary to search one below the truncated value.

On my laptop, this reduces part1() from 2.3us to 2.0us; and part2() from 1.4us to 1.0us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day07.rs
+++ b/src/year2021/day07.rs
@@ -1,11 +1,14 @@
 //! # The Treachery of Whales
 //!
 //! Part one is a disguised definition of the mathematical [median](https://en.wikipedia.org/wiki/Median).
-//! We can calculate the result immediately using the standard algorithm.
+//! We can calculate the result immediately using the standard algorithm. Even though there
+//! are an even number of crabs, any integer between the 500th and 501st crab inclusive will
+//! work (the extra fuel spent by half the crabs perfectly cancels the fuel saved by the other
+//! half, when switching between integers in that range).
 //!
 //! Part two is found by using the [mean](https://en.wikipedia.org/wiki/Mean).
 //! However, since this could be a floating point value and we are using integers we need to check
-//! 3 values total, the rounded result and one value on either side to ensure the correct answer.
+//! both the floor and the ceiling of that result to ensure the correct answer.
 use crate::util::parse::*;
 
 pub fn parse(input: &str) -> Vec<i32> {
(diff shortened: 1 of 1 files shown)


ebblake:
Further enhancement beyond #92

src/year2021/day07.rs
```diff
@@ -24,18 +24,24 @@ pub fn part2(input: &[i32]) -> i32 {
         (n * (n + 1)) / 2
     };
 
-    (-1..=1).map(|delta| input.iter().map(|&x| triangle(x, mean + delta)).sum()).min().unwrap()
+    (0..=1).map(|delta| input.iter().map(|&x| triangle(x, mean + delta)).sum()).min().unwrap()
```
maneatingape, line 30: ❤️

src/year2021/day07.rs
```diff
@@ -24,18 +24,24 @@ pub fn part2(input: &[i32]) -> i32 {
...
 fn median(input: &[i32]) -> i32 {
-    let mut crabs = input.to_vec();
-    let mid = crabs.len() / 2;
-    assert!(crabs.len().is_multiple_of(2));
-
-    crabs.select_nth_unstable(mid);
-    let upper = crabs[mid];
-    let lower = crabs[..mid].iter().max().unwrap();
-    (lower + upper) / 2
+    // The data is dense enough that a radix sort outperforms select_nth_unstable.
```
maneatingape: Using your observeration then simplifying this to just:

```
fn median(input: &[i32]) -> i32 {
    let mut crabs = input.to_vec();
    let middle = crabs.len() / 2;
    *crabs.select_nth_unstable(middle).1
}
```

gives nearly the same speedup as the radix sort, in a much simpler way.

maneatingape (changes requested):

src/year2021/day07.rs
ebblake: Fair enough. I'll respin this to fewer lines of code (back to the quickselect rather than radix sort, but no longer averaging neighbors). Today's [reddit thread](https://www.reddit.com/r/adventofcode/comments/1upmd3r/comment/ow294zb/) has a discussion on levels of effort; askalski did the radix sort because it then let him solve both part 1 and part 2 with O(1) additional effort beyond the creation of two prefix sum arrays (one for counts, one for distances).  We're at the point where 5000 vs. 6000 operations is going to be in the noise compared to the parse, so shorter code starts to mean more than squeezing out every last CPU cycle.

ebblake: My laptop is sill clocking the select_nth_unstable variant closer to 2.0us vs. radix sort at 1.0us (the savings from 2.4us comes from no longer averaging two values). I'm not too fussed, since parse() is still dominating with over 4.0us.

maneatingape (approved):
👍

PR #97, opened 2026-07-10
Add an A* search that outperforms Dijkstra

ebblake:
## Description

For this search problem (optimal path across diagonal endpoints of a square grid of integers 1-9), a naive A* search using Manhattan distance to the goal underestimates. That's because the average tile is 5, but Manhattan distance only accounts for 1; and we end up spending more time on the heuristic than we gain in path pruning (in instrumented code, Dijkstra visits all 250000 nodes of my part 2, while the naive A* is able to prune about 20 of those nodes).  In the other direction, a heuristic built up only from cells to the right or below the current cell overestimates (if the heuristic does not account for the possibility of the lowest-risk path sometimes needing to go up or left through a valley of 1s to skirt around a wall of 9s, then A* will accidentally prune that path despite it being the best).

However, the megathread had an observation on a way to build a deterministic heuristic that never overestimates, while still approaching within 2% of the actual cost - for every diagonal of the grid, reduce a given node's heuristic if it can reach another cheaper node on the same diagonal via an unseen Manhattan path assumed to be all 1s.  Without actually having to visit nodes off the diagonal to see if the assumption is valid, this will sometimes underestimate (so we DO need the A* search as a followup pass, rather than being able to solve the puzzle just by computing heuristics), but never overestimate.  The resulting heuristic is admissible but not consistent; on my input, I had several instances where the estimate reduction applied by looking along the same diagonal resulted in revisiting an earlier bucket that had already been emptied; however, the range of buckets populated is still fairly tight (never more than -1 to 18 away from the current bucket), and the bucket window gradually shifts as the underestimate grows closer to reality the closer we get to the target (for my puzzle, the estimate at the origin was low by only 64 out of nearly 3000).  With the heuristic in place, the A* search drops down to about 55k nodes searched (ie. pruning away over 75% of the grid). The time spent visiting all 250000 nodes to populate the heuristic (with its regular memory accesses that can be vectorized) is less than the time saved by not searching the bulk of the grid (where the search frontier tends to be more random-access and thus slower per iteration), making this a net win over Dijkstra.

On my laptop and input, runtime improves from 155us for part1() and 4.4ms for part2(), to instead being 70us and 3.0ms.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day15.rs
+++ b/src/year2021/day15.rs
@@ -1,18 +1,42 @@
 //! # Chiton
 //!
 //! Traversing a graph with different non-negative edge weights is a job for the classic
-//! [Dijkstra's algorithm](https://www.redblobgames.com/pathfinding/a-star/introduction.html),
+//! [A* algorithm](https://www.redblobgames.com/pathfinding/a-star/introduction.html),
 //! explained really well in the linked blog post.
 //!
-//! To speed things up we use a trick. Classic Dijkstra uses a generic priority queue that
-//! can be implemented in Rust using a [`BinaryHeap`]. However, the total cost follows a strictly
-//! increasing order in a constrained range of values, so we can use a much faster single purpose
-//! data structure instead.
+//! The simplest possible heuristic in A* is to make no estimate at all, making the search
+//! behave the same as Dijkstra's algorithm. With that heuristic, the search frontier visits
+//! nearly every tile, since only a few tiles near the target might have a total risk higher
+//! than the target itself. Slightly better is a heuristic of the Manhattan distance to the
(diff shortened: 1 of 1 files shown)


ebblake:
Learned from https://www.reddit.com/r/adventofcode/comments/rgqzt5/comment/honpqzq/

maneatingape (approved):
Nice heuristic!

Interestingly, this barely moved the needle on my own input (2.3 => 2.2ms) but halved the time for another input I had available (2.3 => 1.2ms).

PR #101, opened 2026-07-20
Use 1-based rather than 0-based binary array

ebblake:
## Description

When storing an implicit binary tree in an array, there is a different formula for the parent when root is 0 than when root is 1.

Aligning levels of the tree to cache-line boundaries and using fewer +1/-1 adjustments when moving between levels of the tree is slightly more efficient.  On my machine, this change improves runtime by about 2.5%, from 730us to 710us.

## Type of change

- [x] Performance improvement
- [X] Bug fix (docs)
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day18.rs
+++ b/src/year2021/day18.rs
@@ -25,7 +25,7 @@
 //! This means that we can store each snailfish number as an implicit data structure in a fixed-size
 //! array. This is faster, smaller and more convenient than using a traditional struct with pointers.
 //! The root node is stored at index 0. For a node at index `i` its left child is at index
-//! `2i + 1`, right child at index `2i + 2` and parent at index `i / 2`. As leaf nodes are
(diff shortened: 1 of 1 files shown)


ebblake:
I have not yet tested whether using a 1-based tree (array of 64 instead of 63 elements, entry 0 unused) would result in enough simplifications to the rest of the math to make the solution run slightly faster (among other effects, it would put all level 5 leaves in the same cacheline, rather than straddling the first separately from the rest).

ebblake:
> I have not yet tested whether using a 1-based tree (array of 64 instead of 63 elements, entry 0 unused) would result in enough simplifications to the rest of the math to make the solution run slightly faster (among other effects, it would put all level 5 leaves in the same cacheline, rather than straddling the first separately from the rest).

I've now tested, and was slightly surprised that the results, while small, were still definitive at being slightly faster on my machine.

maneatingape (approved):
👍  Cleans up the code a bit and slightly faster

PR #102, opened 2026-07-22
Speed up part 1, and demonstrate a compile-time solution to part 2

ebblake:
## Description

The dynamic solution only fills up about 30% of the 44100 cache entries; but it is not much harder to turn the cache into a deterministic dynamic-programming table filled up to 100% at compile time, at which point the runtime for part 2 drops to a mere nanosecond for a table lookup covering all possible inputs (quite a bit faster than the 800us it took beforehand).  Part 1 can also be speed up by recognizing that the visitation pattern is periodic, and thus we can skip forward a period at a time (dropping from 1300 to 100 ns on my machine).  However, since compile time is drastically increased (over 10 seconds on my machine), and since part 1 runtime is in the noise compared to part 2, I suspect that this patch is more of a cool demo, and not something to actually be applied.

## Type of change

- [X] Performance improvement (demo)
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day21.rs
+++ b/src/year2021/day21.rs
@@ -20,10 +20,27 @@ pub fn parse(input: &str) -> State {
 /// The initial deterministic dice roll total is 6 (1 + 2 + 3) and increases by 9 each turn.
 /// An interesting observation is that since the player's position is always modulo 10, we can
 /// also increase the dice total modulo 10, as (a + b) % 10 = (a % 10) + (b % 10).
+/// Additionally, both players end up in the same position every 10 moves, so we can pre-compute
+/// the score per 10 moves before simulating the remainder.
 pub fn part1(input: &State) -> usize {
     let mut state = *input;
     let mut dice = 6;
-    let mut rolls = 0;
+    let ((player_position, _), (other_position, _)) = state;
+
+    // Player 1 loops through offsets 6, 10, 2, 2, 10, 6, 10, 2, 2, 10.
+    let skip_one = 4 * ((player_position + 2) % 10 + 1)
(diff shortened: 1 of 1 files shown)


ebblake:
Done because of https://www.reddit.com/r/adventofcode/comments/1v2fkm5/comment/oz2dkt9/

Part 1 could also be done at compile time (with less effort than the part 2 compilation effects), although the runtime speedup is not as dramatic.

maneatingape:
So I think you know what I'm going to say about part two... 🙂 

I like the clever part one shortcut. The code can be simplified to something like:

```
    let skip = |position, offsets: &[usize]| {
        offsets.iter().map(|offset| (position + offset) % 10 + 1).sum()
    };
    let skip_one = skip(other_position, &[6, 10, 2, 2, 10, 6, 10, 2, 2, 10]);
    let skip_two = skip(other_position, &[5, 8, 9, 8, 5, 10, 3, 4, 3, 10]);
 ```

ebblake:
> So I think you know what I'm going to say about part two... 🙂

Yep! But hey, I had fun seeing how slow I can make the Rust compiler (I was honestly surprised I didn't have to whip out `#[allow(long-running-const-eval)]` as suggested to me by AI).

> 
> I like the clever part one shortcut. The code can be simplified to something like:

I like your compact respelling of the part 1 changes. I'll spin that into a separate patch then close this one, so that this one remains linkable as the demo of compile-time solution. And maybe I can still find ways to shave time off part 2 without abusing the compiler (since the part 1 tweak is in the noise compared to part 2 runtime).

PR #104, opened 2026-07-22
Exploit periodic pattern of deterministic dice

ebblake:
## Description

Part 1 can be sped up by recognizing that 10 slots on a board coupled with deterministic rolls of a d100 dice produce a periodic visitation pattern.  The first player has a cycle of 5, and the second a cycle of 10; so we can compute the overall impact to the score and advance 10 moves at a time until finally running the last few rounds manually to correctly determine the winner.

In part 2, it is faster to use subtraction than modulus for a value that we know is no greater than 18, but where the compiler does not see that range.

On my laptop, this speeds up part1 from 1300ns to 70ns, and part2 from 760us to 700us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day21.rs
+++ b/src/year2021/day21.rs
@@ -19,11 +19,24 @@ pub fn parse(input: &str) -> State {
 
 /// The initial deterministic dice roll total is 6 (1 + 2 + 3) and increases by 9 each turn.
 /// An interesting observation is that since the player's position is always modulo 10, we can
-/// also increase the dice total modulo 10, as (a + b) % 10 = (a % 10) + (b % 10).
+/// also increase the dice total modulo 10, as (a + b) % 10 = (a % 10) + (b % 10). Additionally,
+/// both players end up back in the same position every 10 moves, so we can compute the score per
+/// batch of 10 moves before simulating only the remainder.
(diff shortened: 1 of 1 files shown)


ebblake:
Salvaged from #102

src/year2021/day21.rs
```diff
@@ -67,7 +80,8 @@ fn dirac(state: State, cache: &mut [Option<Pair>]) -> Pair {
     }
 
     let helper = |(win, lose), &(dice, frequency)| {
-        let next_position = (player_position + dice) % 10;
+        let advance = player_position + dice;
+        let next_position = if advance >= 10 { advance - 10 } else { advance };
```
maneatingape, line 84: ❤️ This gives a 10% speedup on part 2.

maneatingape (approved):

src/year2021/day21.rs
ebblake: Modulus of a range-limited value is surprisingly slow - even when the compiler can force it to a multiplication rather than a hardware division, it still chews up more clock cycles than a conditional subtraction.

PR #105, opened 2026-07-31
Pre-compute initial explode actions

ebblake:
## Description

During part two, we are computing a pairing between every two lines. Any line that had one or more depth 4 number pairs will cause the same set of explode actions to be taken for each of the other 99 lines it is paired with, before we get into the main effort of add() in processing split actions.  Rather than repeat early work that will produce the same answer, we can instead pre-compute these initial explode calls during parse time.

On my input, adding a simple counter shows that this reduces the number of calls to explode() from 180k down to 131k. parse() and part1() runtimes are about the same, while part2() speeds up from 600us to 560us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2021/day18.rs
+++ b/src/year2021/day18.rs
@@ -24,9 +24,25 @@
 //!
 //! This means that we can store each snailfish number as an implicit data structure in a fixed-size
 //! array. This is faster, smaller and more convenient than using a traditional struct with pointers.
-//! The root node is stored at index 1 (index 0 is unused). For a node at index `i` its left child
-//! is at index `2i`, right child at index `2i + 1` and parent at index `i / 2`. As leaf nodes are
-//! always greater than or equal to zero, `-1` is used as a special sentinel value for non-leaf nodes.
+//! The root node is stored at index 1 (index 0 is unused by the tree, but see below). For a
+//! node at index `i` its left child is at index `2i`, right child at index `2i + 1` and parent
+//! at index `i / 2`. As leaf nodes are always greater than or equal to zero, `-1` is used as a
+//! special sentinel value for non-leaf nodes.
+//!
+//! Another optimization is realizing that all of the explode actions before the first split can
+//! be pre-computed. Instead of parsing a line `[1,[2,[3,[4,5]]]]` as written, we instead parse it
+//! it as if it had been `[[1,[2,[3,[4,5]]]],0]`, then perform the initial explode actions for that
+//! line up front, such that node 3 contains the value that must be added to the first leaf of a
+//! right-hand value in summation. Node 0 is not used by the implicit tree structure, so we instead
+//! use it as a tri-state value:
+//!
+//! - -2 means this Snailfish number is the result of a sum, for the left side during part one.
+//!   When passed to `add()`, we must shuffle contents one level lower.
+//! - -1 means this Snailfish number was just parsed, but did not explode left. When used on the
+//!   right side of add, any spillover from the left is added to the first leaf on the right.
+//! - Non-negative means this Snailfish number was just parsed, and had an explode that spilled
+//!   left. When used as the right side of an add, any spill from the left is combined with
+//!   this value, then added to the last leaf on the left.
 use crate::util::parse::*;
 use crate::util::thread::*;
 
@@ -49,8 +65,12 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
     input
         .lines()
         .map(|line: &str| {
+            // Treat the line as if it had been `[line,0]`, then perform explode until it is back
+            // at depth 4. This allows later add() operations to do less work. Index 0 and 3
+            // then track the amount spilled left or right from those explodes.
             let mut tree = [-1; 64];
-            let mut i = 1;
+            tree[3] = 0;
+            let mut i = 2;
 
             for b in line.bytes() {
                 match b {
@@ -58,6 +78,11 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
                     b',' => i += 1,
                     b']' => i /= 2,
                     b => tree[i] = b.to_decimal() as i32,
+                }
+            }
+            for pair in (32..48).step_by(2) {
+                if tree[pair] >= 0 {
+                    explode(&mut tree, pair);
                 }
             }
 
@@ -100,29 +125,56 @@ fn worker(iter: ParIter<'_, (&Snailfish, &Snailfish)>) -> Option<i32> {
 /// The initial step creates a new root node then makes the numbers the left and right children
 /// of this new root node, by copying the respective ranges of the implicit trees.
 ///
-/// We can optimize the rules a little. This initial combination is the only time that more than one
-/// pair will be 4 levels deep simultaneously, so we can sweep from left to right on all possible
-/// leaf nodes in one pass.
+/// We can optimize the rules a little. The parse step already ensured that there are no pairs
+/// deeper than 4 levels, and precomputed any explode values to spill between the two halves
+/// of the joined value. All that remains is checking for splits, where each split also takes
+/// care of any additional explodes needed.
 fn add(left: &Snailfish, right: &Snailfish) -> Snailfish {
     let mut tree = [-1; 64];
 
-    tree[4..6].copy_from_slice(&left[2..4]);
-    tree[8..12].copy_from_slice(&left[4..8]);
-    tree[16..24].copy_from_slice(&left[8..16]);
-    tree[32..48].copy_from_slice(&left[16..32]);
+    if left[0] == -2 {
+        // Left comes from a running sum during part one. We need to increase the depth, which
+        // in turn might cause some depth 5 leaves that need explode.
+        tree[3] = 0;
+        tree[4..6].copy_from_slice(&left[2..4]);
+        tree[8..12].copy_from_slice(&left[4..8]);
+        tree[16..24].copy_from_slice(&left[8..16]);
+        tree[32..48].copy_from_slice(&left[16..32]);
 
-    tree[6..8].copy_from_slice(&right[2..4]);
-    tree[12..16].copy_from_slice(&right[4..8]);
-    tree[24..32].copy_from_slice(&right[8..16]);
-    tree[48..64].copy_from_slice(&right[16..32]);
-
-    for pair in (32..64).step_by(2) {
-        if tree[pair] >= 0 {
-            explode(&mut tree, pair);
+        for pair in (32..48).step_by(2) {
+            if tree[pair] >= 0 {
+                explode(&mut tree, pair);
+            }
         }
+    } else {
+        // We are adding two just-parsed numbers; the left is already rooted at 2 and has no depth 5
+        // leaves, making it ready to copy into place.
+        tree[3..24].copy_from_slice(&left[3..24]);
     }
 
+    // Copy the right into place. This value is always just-parsed, with no depth 5 leaves.
+    tree[6..8].copy_from_slice(&right[4..6]);
+    tree[12..16].copy_from_slice(&right[8..12]);
+    tree[24..32].copy_from_slice(&right[16..24]);
+
+    // Adjust by the explode spillover between sides. We ensured that tree[3] contains any
+    // value to spill right, but must check right[0] to see if that sum then spills back left.
+    let (mut i, spill) = if right[0] == -1 { (24, tree[3]) } else { (23, tree[3] + right[0]) };
+    loop {
+        if tree[i] >= 0 {
+            tree[i] += spill;
+            break;
+        }
+        i /= 2;
+    }
+    tree[3] = -1;
+
+    // Now we process all split operations; any further explode actions are done during any split
+    // that creates a temporary depth 5.
     while split(&mut tree) {}
+
+    // Mark this tree as a sum before returning it.
+    tree[0] = -2;
     tree
 }
 
@@ -145,6 +197,9 @@ fn explode(tree: &mut Snailfish, pair: usize) {
             }
             i /= 2;
         }
+    } else {
+        // Store the left spill-out for later use by add().
+        tree[0] = tree[pair];
     }
 
     if pair < 62 {
--- a/src/year2021/day18.rs
+++ b/src/year2021/day18.rs
@@ -180,7 +180,6 @@ fn augment_leaf(tree: &mut [i32], value: i32, mut to: usize) {
     while tree[to] == -1 {
         to /= 2;
     }
-    assert!(to != 0);
     tree[to] += value;
 }
 
--- a/src/year2021/day18.rs
+++ b/src/year2021/day18.rs
@@ -17,10 +17,11 @@
 //!
 //! The addition rules have an important consequence. Exploding removes two leaf nodes at depth 5
 //! and moves them to neighboring nodes. Since exploding repeatedly happens before splitting until
-//! there are no more values at depth 5 this means that the tree will never exceed a depth of 5.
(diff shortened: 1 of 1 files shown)


ebblake:
Inspired by https://www.reddit.com/r/adventofcode/comments/1uzta0a/comment/p0umy44/

ebblake:
It would also be an interesting experiment (one I still need to run) to see if there is any reliable relation between magnitude(reduce([left,0])) and the eventual magnitude(reduce([left,right])) winner. For example, if the winning magnitude always depends on one of the top-ten magnitude values out of the 100 inputs being on the left, we could avoid a full 100*99 add() calls in part 2 and instead focus on the ones most likely to produce a large value.  But it is complicated by the fact that split() means the mapping is not straightforward: `[[0,[0,[0,0]]],[[[9,0],0],0]]` has magnitude 486, but if any explode resulted in the intermediate `[[0,[0,[0,0]]],[[[10,0],0],0]]` that reduces to `[[0,[0,[0,5]]],[[[0,5],0],0]]` with magnitude 300.

src/year2021/day18.rs
```diff
@@ -23,23 +23,38 @@
...
+//! special sentinel value for non-leaf nodes.
+//!
+//! Another optimization is realizing that all of the explode actions before the first split can
+//! be pre-computed. Instead of parsing a line `[1,[2,[3,[4,5]]]]` as written, we instead parse it
+//! it as if it had been `[[1,[2,[3,[4,5]]]],0]`, then perform the initial explode actions for that
+//! line up front, such that node 3 contains the value that must be added to the first leaf of a
+//! right-hand value in summation. Node 0 is not used by the implicit tree structure, so we instead
+//! use it as a tri-state value:
+//!
+//! - -2 means this Snailfish number is the result of a sum, for the left side during part one.
```
maneatingape: Expressing this in types will be clearer, as we really have 2 things (consider all names a loosely held WIP 🙂 ):

Normal uncompressed snailfish number up to 8 pairs, 16 digits, 32 tree size, e.g

`struct Snailfish { tree: [i32; 32] }`

and the compressed version that enables work to be reused:

`struct Compressed { left_spill: i32, right_spill: i32, tree: [i32; 16] }`

maneatingape (changes requested):
I like this optimization!

Tried a few alternative solutions, vec of (digit, depth) pairs and a fixed array with pairs of (digits, skip size) but the tree approach was still faster.

src/year2021/day18.rs
maneatingape: This will shake out the slightly different part 1 / part2 behaviour in the function signatures, e.g.

* `fn add(left: &Compressed, right: &Compressed) -> Snailfish`
* `fn explode(tree: &Snailfish) -> Compressed {`
* `fn magnitude(tree: Snailfish) -> i32`

and probably a new function e.g.

* `fn compress(tree: &Snailfish) -> Compressed`

ebblake: Patch updated. I also got another 40us boost when tweaking how part2() parallizes the list of Compressed pairings.

src/year2021/day18.rs
```diff
@@ -61,72 +76,119 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
+    let mut snailfish = accum.unwrap();
+    magnitude(&mut snailfish)
 }
 
-/// Pair addition is independent so we can parallelize across multiple threads.
-fn worker(iter: ParIter<'_, (&Snailfish, &Snailfish)>) -> Option<i32> {
-    iter.map(|&(a, b)| magnitude(&mut add(a, b))).max()
+/// Find the largest magnitude of any two snailfish numbers, remembering that snailfish addition
+/// is *not* commutative.
+pub fn part2(input: &[Compressed]) -> i32 {
```
maneatingape, line 106: Careful here, this is working by _accident_. If you re-order the numbers this breaks.

The parallel iterator does not guarantee any ordering, so the `enumerate` is based of the _thread local_ ordering, not the global one.

Can be fixed with `let items: Vec<_> = input.iter().enumerate().collect()`. This is only 100 elements instead of 10,000 so results in the same nice speedup.

maneatingape (changes requested):
Detailed review in progress, one quick fix first...

Eliminating the 10,000 element vec is a nice touch, just be aware of concurrency gotchas.

src/year2021/day18.rs
ebblake: Fixed.

ebblake: Should the enumerate be done during parse, or is it fine to keep that copy at the front of parse2()?

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
+    // Build a master list to share among threads.
+    let items: Vec<_> = input.iter().enumerate().collect();
 
     // Use as many cores as possible to parallelize the calculation.
-    let result = spawn_parallel_iterator(&pairs, worker);
-    result.into_iter().flatten().max().unwrap()
-}
+    let result = spawn_parallel_iterator(&items, |iter| {
+        iter.map(|(i, a)| {
```
maneatingape: Change this to `flat_map` then the inner loop can be simpified to

```
            input
                .iter()
                .enumerate()
                .filter(move |&(j, _)| i != j)     <-- Need to add move here
                .map(|(_, b)| magnitude(add(a, b)))
```

src/year2021/day18.rs
```diff
@@ -38,15 +46,22 @@ const IN_ORDER: [usize; 30] = [
     29, 15, 30, 31,
 ];
 
-type Snailfish = [i32; 64];
+type Snailfish = [i16; 64];
```
maneatingape: * Non-native widths (other than 32 or 64 bits) need expensive conversion, so I would avoid premature optimization unless benchmarking shows a definitive improvement.
* On the flip side, since we're compressing numbers before addition (3 levels) we'll never go over 4 levels so this can be cut in half to 32 elements long
* I'd recommend the [new type idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to make this symmetrical to the Compressed struct

Taken together this would look like:

```
#[derive(Clone, Copy, Default)]
struct Snailfish {
    tree: [i32; 32],
}
```

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
+    // that creates a temporary depth 5.
     while split(&mut tree) {}
+
     tree
 }
 
+/// Perform all initial explodes to create a compressed number from a snailfish number.
+/// The incoming number must not have any depth 5 members. This is a destructive operation, as no
+/// caller needs the original afterwards.
+fn compress(tree: &mut Snailfish) -> Compressed {
```
maneatingape: Take ownership of the input e.g. `fn compress(full: Snailfish) -> Compressed {`
then we can re-use the same array, compressing in place, using only the first 16 elements of the array.

This will eliminate the internal copies.

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
         })
         .collect()
 }
 
 /// Add all snailfish numbers, reducing to a single magnitude.
-pub fn part1(input: &[Snailfish]) -> i32 {
-    let mut sum = input.iter().copied().reduce(|acc, n| add(&acc, &n)).unwrap();
-    magnitude(&mut sum)
+pub fn part1(input: &[Compressed]) -> i32 {
+    let mut left = input[0];
```
maneatingape: If both `compress` and `manitude` take ownership of the input, and `Snailfish` and `Compressed` are both copy then this can simpify to a `reduce` with side-effects e.g. 

```
    let mut sum = Snailfish::default();

    input
        .iter()
        .copied()
        .reduce(|a, b| {
            sum = add(&a, &b);
            compress(sum)
        })
        .unwrap();

    magnitude(sum)
```

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
 /// The initial step creates a new root node then makes the numbers the left and right children
 /// of this new root node, by copying the respective ranges of the implicit trees.
 ///
-/// We can optimize the rules a little. This initial combination is the only time that more than one
-/// pair will be 4 levels deep simultaneously, so we can sweep from left to right on all possible
-/// leaf nodes in one pass.
-fn add(left: &Snailfish, right: &Snailfish) -> Snailfish {
+/// We can optimize the rules a little; the first round of explode was already done in creating
+/// compressed arguments, and a split runs its own inline explode.
+fn add(left: &Compressed, right: &Compressed) -> Snailfish {
```
maneatingape, line 128: If `Compressed` uniformly compressed into the first 16 elements, then this would be symmetrical
e.g.

```
    tree[4..6].copy_from_slice(&left[2..4]);
    tree[8..12].copy_from_slice(&left[4..8]);
    tree[16..24].copy_from_slice(&left[8..16]);

    tree[6..8].copy_from_slice(&right[2..4]);
    tree[12..16].copy_from_slice(&right[4..8]);
    tree[24..32].copy_from_slice(&right[8..16]);
```

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
+    // Copy the right into place, with proper interleaving.
+    tree[6..8].copy_from_slice(&right.tree[4..6]);
+    tree[12..16].copy_from_slice(&right.tree[8..12]);
+    tree[24..32].copy_from_slice(&right.tree[16..24]);
 
-    for pair in (32..64).step_by(2) {
-        if tree[pair] >= 0 {
-            explode(&mut tree, pair);
+    // Adjust by the explode spillover between sides.
+    let (mut i, spill) = if right.left_spill == -1 {
```
maneatingape: Clearer to be explicit e.g.

```
    match (right.left_spill, left.right_spill) {
        (-1, -1) => (),
        (left, -1) => explode(&mut tree, left, 23),
        (-1, right) => explode(&mut tree, right, 24),
        (left, right) => tree[23] += left + right,
    }
```

maneatingape: A nice outcome is that we can simplify the `explode` function e.g.

```
fn explode(tree: &mut [i32], value: i32, mut to: usize) {
    while tree[to] == -1 {
        to /= 2;
    }
    tree[to] += value;
}
```

src/year2021/day18.rs
```diff
@@ -61,72 +76,120 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
+        if tree[pair] >= 0 {
+            explode(tree, pair);
+        }
+    }
+
+    // Copy the results into the compressed tree.
+    let mut compressed = [-1; 24];
+    compressed[4..24].copy_from_slice(&tree[4..24]);
+
+    Compressed { left_spill: tree[0], right_spill: tree[3], tree: compressed }
```
maneatingape: With the internal copies then `left_spill` will always be `tree[16` and `right_spill` always be `tree[31]`.

src/year2021/day18.rs
```diff
@@ -196,5 +262,5 @@ fn magnitude(tree: &mut Snailfish) -> i32 {
             tree[i] = 3 * tree[2 * i] + 2 * tree[2 * i + 1];
         }
     }
-    tree[1]
+    tree[1] as i32
```
maneatingape: Update loop to `for i in (1..16).rev() {`.

The function can also take ownership of the input e.g. somehting like `fn magnitude(full: Snailfish) -> i32 {`

maneatingape (changes requested):
This is shaping up really nicely and I think we can save even more time with some modifications.

src/year2021/day18.rs
ebblake: I don't think add(Snailfish::default, compressed[0]) gives the same result as starting the reduce with add(compressed[0], compressed[1]).

src/year2021/day18.rs
ebblake: 32 elements might work - but it means rewriting explode() to cope correctly when used by split(). Right now, split temporarily increases the depth to 5, then explodes it back to 4; if the struct itself is capped to depth 4, then split must behave differently for a depth 4 item being split (immediately add both halves to the correct neighbors) than for a depth 2 or 3 item being split (replace the depth n item with two depth n+1 items).

src/year2021/day18.rs
ebblake: I guess I'm still new enough at Rust that I'm struggling a bit on when to use references vs ownership transfers; but I agree that an ownership transfer is the best paradigm for this puzzle, so I'm working to refactor it along those lines.

src/year2021/day18.rs
ebblake: Copy-wise, is it better to have Compressed be a 14-element array plus the 2 items for left and right spill, so that the struct itself is 16 elements?  I'll get the updated code working with i32 items, then benchmark whether i16 makes any noticeable difference (cutting memory in half, but making edits to the memory more expensive because of hardware instructions being trickier).

src/year2021/day18.rs
maneatingape: `sum` is immediatley overwritten, so this default value is never used, but it's needed to keep the compiler happy.

ebblake: okay, I see my confusion; I was thinking in terms of fold() not reduce().

src/year2021/day18.rs
ebblake: benchmark completed; i32 outperformed i16, but only barely. I'm keeping it native size.

src/year2021/day18.rs
ebblake: That's only half an explode, so I renamed it augment_leaf().

src/year2021/day18.rs
```diff
@@ -61,124 +80,134 @@ pub fn parse(input: &str) -> Vec<Snailfish> {
...
-    tree[pair + 1] = -1;
-    tree[pair / 2] = 0;
+/// Augment the correct leaf by the given non-negative value. Walks up the tree starting at the
+/// given index until finding a leaf node. Storing the tree as an implicit structure has a nice
+/// benefit that finding the next left or right node is straightforward.
+fn augment_leaf(tree: &mut [i32], value: i32, mut to: usize) {
+    while tree[to] == -1 {
+        to /= 2;
+    }
+    assert!(to != 0);
```
maneatingape: Leftover debugging assert is costing ~10% performance, I'll remove this then merge the PR.

src/year2021/day18.rs
```diff
@@ -17,17 +17,26 @@
...
+//! snailfish tree is 1 + 2 + 4 + 8 + 16 = 2⁵ - 1 = 31 nodes.
 //!
 //! This means that we can store each snailfish number as an implicit data structure in a fixed-size
 //! array. This is faster, smaller and more convenient than using a traditional struct with
 //! pointers. The root node is stored at index 1 (index 0 is unused). For a node at index `i` its
 //! left child is at index `2i`, right child at index `2i + 1` and parent at index `i / 2`. As leaf
 //! nodes are always greater than or equal to zero, `-1` is used as a special sentinel value for
 //! non-leaf nodes.
+//!
+//! Another optimization is realizing that all of the explode actions before the first split can
```
maneatingape, line 33: ❤️

maneatingape (approved):
One small fix, which I'll make then merge.

PR #106, opened 2026-08-03
Prefer u64 over u128

ebblake:
## Description

While Rust can support a bitmap of raw ASCII values, it is decidely less efficient at doing so (1<<96 required determining which of two underlying registers to tweak).  We can exploit that all of the ASCII letters fit within a 64-bit range with a simple mask, and instantly gain some speed.

On my laptop, this improves from 12s per part to 2us per part.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`


ebblake:
Inspired by https://www.reddit.com/r/adventofcode/comments/1ve4s0o/comment/p1gnu80/

maneatingape (approved):
🔥

PR #107, opened 2026-08-13
Use single BFS search for both parts

ebblake:
## Description

Cut the runtime approximately in half by realizing that the search for part two covers 95% of the search for part one.  We can further speed up the search by reusing grid instead of a separate seen array as the witness for an unvisited node.

On my laptop, this cuts runtime from 120us to 59us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2022/day12.rs
+++ b/src/year2022/day12.rs
@@ -12,8 +12,8 @@
 //! For part two we could search for all `a` locations and repeatedly start a BFS search from there,
 //! then find the lowest value. However, a much faster approach is to search *backwards* from the
 //! end location. Due to the fact that BFS always explores closest nodes first this will find the
-//! closest `a` location in a single search. For part one it will have the same result, so we
-//! can reuse the same code.
+//! closest `a` location in a single search. In fact, we can just run one single search, finding
+//! the part two answer first, then continuing on to the `S` location for part one.
 //!
 //! [`Grid`]: crate::util::grid
 //! [`Point`]: crate::util::point
@@ -22,42 +22,35 @@ use std::collections::VecDeque;
 use crate::util::grid::*;
 use crate::util::point::*;
 
-type Input = (Grid<u8>, Point);
+type Input = (u32, u32);
 
 /// Uses the utility [`Grid`] module to parse a 2D array of ASCII characters.
 ///
 /// [`Grid`]: crate::util::grid
 pub fn parse(input: &str) -> Input {
-    let grid = Grid::parse(input);
+    let mut grid = Grid::parse(input);
+
+    // Run the BFS algorithm implementation with the reversed height transition rules baked in.
+    // In fact, we don't need a separate seen grid; we can modify the original grid in place.
     let start = grid.find(b'E').unwrap();
-    (grid, start)
-}
+    let mut todo = VecDeque::from([(start, b'z' - 1, 1)]);
+    grid[start] = 0;
+    let mut part_two = None;
 
-/// Find the shortest path from `E` to `S`.
-pub fn part1(input: &Input) -> u32 {
-    bfs(input, b'S')
-}
-
-/// Find the shortest path from `E` to closest `a`.
-pub fn part2(input: &Input) -> u32 {
-    bfs(input, b'a')
-}
-
-/// BFS algorithm implementation with the reversed height transition rules baked in.
-fn bfs(input: &Input, end: u8) -> u32 {
-    let (grid, start) = input;
-    let mut todo = VecDeque::from([(*start, 0)]);
-    let mut seen = grid.same_size_with(false);
-    seen[*start] = true;
-
-    while let Some((point, cost)) = todo.pop_front() {
-        if grid[point] == end {
-            return cost;
-        }
+    while let Some((point, height, cost)) = todo.pop_front() {
         for next in ORTHOGONAL.map(|d| d + point) {
-            if grid.contains(next) && !seen[next] && height(grid, point) - height(grid, next) <= 1 {
-                todo.push_back((next, cost + 1));
-                seen[next] = true;
+            if !grid.contains(next) {
+                continue;
+            }
+            if grid[next] == b'S' {
+                return (cost, part_two.unwrap());
+            }
+            if grid[next] >= height {
+                if grid[next] == b'a' {
+                    part_two = part_two.or(Some(cost));
+                }
+                todo.push_back((next, grid[next] - 1, cost + 1));
+                grid[next] = 0;
             }
         }
     }
@@ -65,11 +58,12 @@ fn bfs(input: &Input, end: u8) -> u32 {
     unreachable!()
 }
 
-/// Map `S` to `a` and `E` to `z`, otherwise use the value unchanged.
-fn height(grid: &Grid<u8>, point: Point) -> i32 {
-    match grid[point] {
-        b'S' => b'a' as i32,
-        b'E' => b'z' as i32,
-        b => b as i32,
-    }
+/// Find the shortest path from `E` to `S`.
+pub fn part1(input: &Input) -> u32 {
+    input.0
+}
+
+/// Find the shortest path from `E` to closest `a`.
+pub fn part2(input: &Input) -> u32 {
+    input.1
 }
--- a/src/year2022/day12.rs
+++ b/src/year2022/day12.rs
@@ -39,18 +39,17 @@ pub fn parse(input: &str) -> Input {
 
     while let Some((point, height, cost)) = todo.pop_front() {
         for next in ORTHOGONAL.map(|d| d + point) {
-            if !grid.contains(next) {
-                continue;
-            }
-            if grid[next] == b'S' {
-                return (cost, part_two.unwrap());
-            }
-            if grid[next] >= height {
-                if grid[next] == b'a' {
-                    part_two = part_two.or(Some(cost));
+            if grid.contains(next) {
+                if grid[next] == b'S' {
+                    return (cost, part_two.unwrap());
                 }
-                todo.push_back((next, grid[next] - 1, cost + 1));
-                grid[next] = 0;
+                if grid[next] >= height {
+                    if grid[next] == b'a' {
+                        part_two = part_two.or(Some(cost));
+                    }
+                    todo.push_back((next, grid[next] - 1, cost + 1));
+                    grid[next] = 0;
+                }
             }
         }
     }
--- a/tests/year2022/day12.rs
+++ b/tests/year2022/day12.rs
@@ -5,8 +5,7 @@ Sabqponm
 abcryxxl
 accszExk
 acctuvwj
-abdefghi
-";
+abdefghi";
 
 #[test]
 fn part1_test() {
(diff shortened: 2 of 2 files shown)


src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
-                seen[next] = true;
+            if grid.contains(next) {
+                let next_height = height(&grid, next);
+                if curr_height - next_height <= 1 {
+                    if let Some(part_two) = first_a {
+                        if grid[next] == b'S' {
+                            return (cost, part_two);
+                        }
+                    } else if grid[next] == b'a' {
+                        first_a = Some(cost);
```
ebblake: This diff did not make much of a performance difference, but allows for fewer nested if. I guess that means the compiler is already doing a decent job of optimizing the conditions.
```
diff --git i/src/year2022/day12.rs w/src/year2022/day12.rs
index d0466bd..b3b70c4 100644
--- i/src/year2022/day12.rs
+++ w/src/year2022/day12.rs
@@ -33,7 +33,7 @@ pub fn parse(input: &str) -> Input {
     // Run the BFS algorithm implementation with the reversed height transition rules baked in.
     // In fact, we don't need a separate seen grid; we can modify the original grid in place.
     let mut todo = VecDeque::from([(start, height(&grid, start), 1)]);
-    let mut first_a = None;
+    let mut part_two = u32::MAX;
     grid[start] = 0;

     while let Some((point, curr_height, cost)) = todo.pop_front() {
@@ -41,12 +41,10 @@ pub fn parse(input: &str) -> Input {
             if grid.contains(next) {
                 let next_height = height(&grid, next);
                 if curr_height - next_height <= 1 {
-                    if let Some(part_two) = first_a {
-                        if grid[next] == b'S' {
-                            return (cost, part_two);
-                        }
+                    if grid[next] == b'S' {
+                        return (cost, part_two);
                     } else if grid[next] == b'a' {
-                        first_a = Some(cost);
+                        part_two = part_two.min(cost);
                     }
                     todo.push_back((next, next_height, cost + 1));
                     grid[next] = 0;
```

src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
-/// Find the shortest path from `E` to `S`.
-pub fn part1(input: &Input) -> u32 {
-    bfs(input, b'S')
-}
-
-/// Find the shortest path from `E` to closest `a`.
-pub fn part2(input: &Input) -> u32 {
-    bfs(input, b'a')
-}
+    // Run the BFS algorithm implementation with the reversed height transition rules baked in.
```
maneatingape, line 33: Modifying the grid in place, enables a few tricks that can speed this up even more.

First, find the end point at 'S'
```
let end = grid.find(b'S').unwrap();
```

Next, erase the 2 special characters 'S' and 'E'
```
grid[start] = 0;
grid[end] = b'a';
```

src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
+    // In fact, we don't need a separate seen grid; we can modify the original grid in place.
+    let mut todo = VecDeque::from([(start, height(&grid, start), 1)]);
```
maneatingape: Now we can remove the `height` method entirely and this line simplified to
```
let mut todo = VecDeque::from([(start, b'z', 1)]);
```

src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
+    let mut first_a = None;
```
maneatingape: Treating both endpoints equally was slightly faster, e.g.
```
let mut part_one = None;
let mut part_two = None;
```

src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
-    while let Some((point, cost)) = todo.pop_front() {
-        if grid[point] == end {
-            return cost;
-        }
+    while let Some((point, curr_height, cost)) = todo.pop_front() {
         for next in ORTHOGONAL.map(|d| d + point) {
-            if grid.contains(next) && !seen[next] && height(grid, point) - height(grid, next) <= 1 {
-                todo.push_back((next, cost + 1));
-                seen[next] = true;
+            if grid.contains(next) {
```
maneatingape, line 42: We can now combine the checks into a single if and remove a layer of nesting
```
if grid.contains(next) && i32::from(curr_height) - i32::from(grid[next]) <= 1 {
```

src/year2022/day12.rs
```diff
@@ -5,66 +5,69 @@
...
+                let next_height = height(&grid, next);
+                if curr_height - next_height <= 1 {
+                    if let Some(part_two) = first_a {
```
maneatingape: Treating part one and two equally was slightly faster e.g.
```
                if next == end {
                    part_one = part_one.or(Some(cost));
                    if part_two.is_some() {
                        return (part_one, part_two);
                    }
                }
                else if grid[next] == b'a' {
                    part_two = part_two.or(Some(cost));
                    if part_one.is_some() {
                        return (part_one, part_two);
                    }
                }
```

maneatingape (changes requested):
Nice speedup. Building on the idea of mutating the grid in place, we can go ~25% faster!

src/year2022/day12.rs
ebblake: With the logic, part_two will ALWAYS be satisfied before part_one. I may still play around with the conditionals to see if any rearranging for short-circuiting operations goes faster than others (for example, I already know that checking is_none prior to == b'a' was faster than checking in the opposite order).

maneatingape: That's fair..trying something like this was a little faster and cleaner:

```
                if grid[next] == b'a' {
                    if next == end {
                        part_one = Some(cost);
                    } else {
                        part_two.get_or_insert(cost);
                    }

                    if let Some(result) = part_one.zip(part_two) {
                        return result;
                    }
                }
```

maneatingape: Using `std::mem::replace` was a little faster, but perhaps too "clever" (e.g. not worth the decrease in clarity)

```
                let cell = replace(&mut grid[next], 0);

                if cell == b'a' {
                    if next == end {
                        part_one = Some(cost);
                    } else {
                        part_two.get_or_insert(cost);
                    }

                    if let Some(result) = part_one.zip(part_two) {
                        return result;
                    }
                }

                todo.push_back((next, cell, cost + 1));
```

ebblake:
Would it be worth repeating the magic in #98 of creating an oversized vector to hold the grid with a known border, so that we can eliminate boundary checks?

src/year2022/day12.rs
ebblake: I'm having a hard time seeing much difference in timings over the various approaches I tried.  But on the grounds that the compiler might be able to better parameterize the code if two blocks look similar and their only difference is the condition gating them, I was about to use your approach, before `just` warned:
```
warning: called `unwrap` on `part_two` after checking its variant with `is_some`
  --> src/year2022/day12.rs:48:52
   |
47 |                     if part_two.is_some() {
   |                     --------------------- help: try: `if let Some(<item>) = part_two`
48 |                         return (part_one.unwrap(), part_two.unwrap());
   |                                                    ^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_unwrap
   = note: `#[warn(clippy::unnecessary_unwrap)]` on by default
```

src/year2022/day12.rs
ebblake: With an extra border for top and bottom, the embedded newlines in the input already serve as left and right borders. I got another speedup from 50us to 40us by eliminating the bounds checks.

src/year2022/day12.rs
ebblake: I got another speedup from 40us to 35us when I hoisted the check for next == end to instead be outside the height subtraction loop as grid[next] == b'S', no need to hunt for end up front.

src/year2022/day12.rs
```diff
@@ -12,64 +12,60 @@
...
-            if grid.contains(next) && !seen[next] && height(grid, point) - height(grid, next) <= 1 {
-                todo.push_back((next, cost + 1));
-                seen[next] = true;
+    while let Some((point, height, cost)) = todo.pop_front() {
+        for delta in [1, -1, width, -width] {
+            let next = point.wrapping_add(delta as usize);
+            if grid[next] == b'S' {
+                return (cost, part_two.unwrap());
+            }
+            if i32::from(height) - i32::from(grid[next]) <= 1 {
```
maneatingape: Experimenting a little, I was able to simplify this to 

`if grid[next] >= height {`

by moving the offset 1 calculation to ` todo.push_back((next, grid[next] - 1, cost + 1));`

while keeping the same performance.

maneatingape (changes requested):
For this problem, I prefer sticking with the slightly slower but clearer Grid utillity class, but otherwise keeping the approach the same.

ebblake:
> For this problem, I prefer sticking with the slightly slower but clearer Grid utillity class, but otherwise keeping the approach the same.

I wonder - would it be worth enhancing the Grid utility to have a parse variant that intentionally provides a newline border on top and bottom while preserving the in-data newline border for both left and right?  Then you can still use the Point traversals without bounds checks on this and several other puzzles.

But I'll go ahead and back out that part of my change plus fold in your suggestion on modifying the height stored into todo.

maneatingape:
> I wonder - would it be worth enhancing the Grid utility to have a parse variant that intentionally provides a newline border on top and bottom while preserving the in-data newline border for both left and right? 

That's an interesting idea, something like `Grid::parse_with_padding(input: &str, padding: u8) -> Self`.
Definitely worth an experiment as there's 12 solutions that reference `contains` and could be faster.

ebblake:
> That's an interesting idea, something like `Grid::parse_width_padding(input: &str, padding: u8) -> Self`. Definitely worth an experiment as there's 12 solutions that reference `contains` and could be faster.

Patch updated along those lines (I independently chose parse_border(input: &str) before seeing your comment), and 2021 day 9 updated to use it. I actually got a slight slowdown there; but that may be because the iteration over x and y to generate distinct Points for every interior may be more expensive than the previous iteration over every integer index of the underlying 1D storage (where iterating on the mid-grid newlines was a nop). The flexibility to specify your own border instead of newline may be overkill (it is more expensive to change out the mid-text newlines).

maneatingape:
Can you split the grid padding change out into its own PR, then we can iterate on it separately. As a rule of thumb before adding a utility method it should be useful in 3+ solutions (which I think is the case, but let's check)

ebblake:
> Can you split the grid padding change out into its own PR, then we can iterate on it separately. As a rule of thumb before adding a utility method it should be useful in 3+ solutions (which I think is the case, but let's check)

Sure. Working on identifying at least one other puzzle that would benefit besides the two I've already touched today.

ebblake:
Padding changes are now fully in #108, and this one just merges part 1 and 2. The end result is that there is still more speedups in the pipeline, but this one is already a sizeable improvement on its own.

maneatingape (approved):
👍

PR #108, opened 2026-08-14
Add new Grid::parse_with_border method

ebblake:
## Description

Several puzzles perform motion along a grid where the motion can be naturally bounded by an explicit newline character. In those cases, discarding the newline in the input and making the code do a .contains() check is more overhead than creating an oversized grid where the input's newlines serve as left AND right boundary (thanks to 1D wraparound), and where we can supply additional newlines on top and bottom at parse time.

While there are probably other puzzles that would also benefit, I picked the following to demonstrate the utility of the new parse_with_border() method, with timings on my laptop and inputs:

 - 2020 day 11
   - Non-SIMD timing improves from part1() 5.3ms and part2() 4.1ms to now complete in 4.3ms and 4.0ms
   - SIMD code was already supplying its own border, but timings with the updated transpose stay at part1() 200us and part2 1.4ms.
 - 2021 day 9 Timing remains around parse() 80us; updating #98 to use the new interface does not hurt the timing.
 - 2022 day 12 Timing improves from part1+part2 56+51us to 51+46us.
 - 2023 day 16 Timing improves from parse() 330us to 325us.
 - 2023 day 21 Timing improves from 310us to 260us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/util/grid.rs
+++ b/src/util/grid.rs
@@ -16,13 +16,18 @@
 //! assert_eq!(grid[point], b'2');
 //! ```
 //!
-//! A convenience [`parse`] method creates a `Grid` directly from a 2-dimensional set of
-//! ASCII characters, a common occurrence in Advent of Code inputs. The [`same_size_with`] function
-//! creates a grid of the same size that can be used in BFS algorithms for tracking visited
-//! locations or for tracking cost in Dijkstra.
+//! Two convenience methods, [`parse`] and [`parse_with_border`], create a `Grid` directly from a
+//! 2-dimensional set of ASCII characters, a common occurrence in Advent of Code inputs. The former
+//! strips all newlines, and [`contains`] is then useful to prevent accidental wraparound between
+//! lines. The latter not only preserves newlines in the input, but adds a row of newlines above and
+//! below, for algorithms where newline serves as a natural barrier without needing to use
+//! [`contains`]. The [`same_size_with`] function creates a grid of the same size that can be used
+//! in BFS algorithms for tracking visited locations or for tracking cost in Dijkstra.
 //!
 //! [`Point`]: crate::util::point
 //! [`parse`]: Grid::parse
+//! [`parse_with_border`]: Grid::parse_with_border
+//! [`contains`]: Grid::contains
 //! [`same_size_with`]: Grid::same_size_with
 use std::ops::{Index, IndexMut};
 
@@ -45,6 +50,25 @@ impl Grid<u8> {
         let bytes = raw.concat();
 
         Self { width, height, bytes }
+    }
+
+    #[must_use]
+    pub fn parse_with_border(input: &str) -> Self {
+        // Size things large enough so that both orthogonal and diagonal access hits a newline.
+        // This shifts 0,0 to 1,1. Non-newline iteration would be `1..height-1` and `1..width`,
+        // although it still often faster to iterate `0..height` and `0..width` when visiting
+        // newline is harmless. For convenience, the allocation is oversized to compensate for unit
+        // tests that omit a trailing newline.
+        let width = input.lines().next().unwrap().len() + 1;
+        let height = input.len().div_ceil(width) + 2;
+        let size = width * height + 1;
+        let mut bytes = Vec::with_capacity(size);
+
+        bytes.resize(width + 1, b'\n');
+        bytes.extend_from_slice(input.as_bytes());
+        bytes.resize(size, b'\n');
+
+        Self { width: width as i32, height: height as i32, bytes }
     }
 
     pub fn print(&self) {
--- a/src/year2020/day11.rs
+++ b/src/year2020/day11.rs
@@ -15,7 +15,8 @@ use crate::util::point::*;
 const SEAT: u8 = b'L';
 
 pub fn parse(input: &str) -> Grid<u8> {
-    Grid::parse(input)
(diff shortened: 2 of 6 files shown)


ebblake:
For my own reference, on 2020 day 11, the code claims
```
        // Input grid is taller than it is wide. To make efficient use of the wide SIMD operations:
        // * Change the existing newline border to empty, but keep it to eliminate bounds checking.
        // * Transpose the input grid to make it wider than it is tall.
````
But I've seen inputs that range from 92x99, 95x98, and 98x90. So some inputs are wider than they are tall. Also, the SIMD code was a bit confusing in that it computes width by rounding to LANE_WIDTH and THEN adding 2, so the just-created grid has a non-power-of-two width.

**Comment by ebblake** [comment]
Conflicts with the current state of #107 ; this one should probably go in first and then I rebase that one.

**Comment by ebblake** [comment]
I was a bit surprised to note that in some of the spots where I tested (such as 2021 day 9), iterating from 0..height and skipping on a newline, rather than switching to 1..height-1 to avoid the skip test on known newlines, could still outperform.  Apparently, Rust is better able to handle loop iterations where EVERY array element is visited, rather than having the iterator skip a spot once per stride.

**Comment by maneatingape** [comment]
> Conflicts with the current state of https://github.com/maneatingape/advent-of-code-rust/pull/107 ; this one should probably go in first and then I rebase that one.

Let's do it the other way around, as #107 is in a good state and ready to merge once the padding changes are removed.

This PR is larger, more speculative and will need some experimentation and benchmarking.

**Comment by ebblake** [comment]
> > Conflicts with the current state of #107 ; this one should probably go in first and then I rebase that one.
> 
> Let's do it the other way around, as #107 is in a good state and ready to merge once the padding changes are removed.

Okay, let me touch up both series then.

> 
> This PR is larger, more speculative and will need some experimentation and benchmarking.

Understood.

File: src/util/grid.rs [hunk 1]
```diff
@@ -47,6 +52,24 @@ impl Grid<u8> {
...
 
+    #[must_use]
+    pub fn parse_with_border(input: &str) -> Self {
+        // Size things large enough so that both orthogonal and diagonal access hits a newline.
+        // This shifts 0,0 to 1,1. Non-newline iteration would be `1..height-1` and `1..width`,
+        // although it still often faster to iterate `0..height` and `0..width` when visiting
+        // newline is harmless.
+        let width = input.lines().next().unwrap().len() + 1;
+        let height = (input.len() / width + 2) as i32;
+        let mut bytes = Vec::with_capacity(input.len() + 2 * width + 1);
```
ebblake: If I size this one byte larger, then the code will work around unit tests that do not include a trailing newline but still want in-bounds access to the diagonal neighbors of the last byte. Input files are immune, but you can see by my changes to 2020/11 and 2022/12 that some of the unit tests triggered runtime out-of-bounds with just the one extra byte given here, compared to the possibility of 2 extra bytes.

ebblake: To make it clearer, for an input file of (x+1)\*y bytes, Grid::parse produces a Vec of x\*y bytes with the data ORIGIN at 0,0, while Grid::parse_with_borders currently produces a grid (x+1)\*(y+1)+1 bytes with the data beginning at 1,1 and every data byte having its diagonal neighbors in the grid, IF the input had a trailing newline. But if the input was (x+1)\*y-1 bytes, supplying yet one more trailing byte here would overcome that, at the expense of (x+1)\*(y+1)+2 bytes for normal inputs. Probably doesn't hurt in the long run to be oversized; while power-of-two allocations can often be faster, many grid puzzles are not already powers of two in either x or y (and even if the grid is square, the newlines make the border parse non-square), and the whole point of oversizing the parse is so that traversals starting from a data point never leave the grid because of the border, so those algorithms won't care about the grid having extra but unreachable memory.

src/util/grid.rs
```diff
@@ -47,6 +52,25 @@ impl Grid<u8> {
...
     }
 
+    #[must_use]
+    pub fn parse_with_border(input: &str) -> Self {
+        // Size things large enough so that both orthogonal and diagonal access hits a newline.
+        // This shifts 0,0 to 1,1. Non-newline iteration would be `1..height-1` and `1..width`,
+        // although it still often faster to iterate `0..height` and `0..width` when visiting
+        // newline is harmless. For convenience, the allocation is oversized to compensate for unit
+        // tests that omit a trailing newline.
+        let width = input.lines().next().unwrap().len() + 1;
```
maneatingape, line 62: Break out the total size explicitly instead of calculating it slightly differently in 2 locations, e.g.

```
        let width = input.lines().next().unwrap().len() + 1;
        let height = input.len().div_ceil(width) + 2;
        let size = width * height + 1;
        let mut bytes = Vec::with_capacity(size);

        bytes.resize(width + 1, b'\n');
        bytes.extend_from_slice(input.as_bytes());
        bytes.resize(size, b'\n');

        Self { width: width as i32, height: height as i32, bytes }
```

maneatingape (changes requested):
Nice addition to the utility `Grid` module

* Faster across the board / equivalent to the hand rolled approach
* Removing the bounds checks simplifies the code
* Applicable to mutliple solutions (including some not in this PR that can be added later)

maneatingape (approved):
👍

PR #109, opened 2026-08-15
Less memory churn in search for candidate lines

ebblake:
## Description

Pre-allocating the FastSets rather than growing them as edges are encountered drastically reduces the runtime.  The sets are almost small enough that an in-memory O(n^2) sort of 1600 comparisons for a 40-line input might still beat the O(n) set intersection with the overhead of hashing.

On my laptop, this cuts runtime from 3.4us to 1.4us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2022/day15.rs
+++ b/src/year2022/day15.rs
@@ -86,10 +86,11 @@ pub fn part2(input: &[Input]) -> u64 {
 /// Of the entire 4000000 by 4000000 area the missing beacon must be located in the only
 /// square area not covered by a scanner.
 pub fn part2_testable(input: &[Input], size: i32) -> u64 {
-    let mut top = FastSet::new();
-    let mut left = FastSet::new();
-    let mut bottom = FastSet::new();
-    let mut right = FastSet::new();
+    let capacity = input.len();
+    let mut top = FastSet::with_capacity(capacity);
+    let mut left = FastSet::with_capacity(capacity);
+    let mut bottom = FastSet::with_capacity(capacity);
+    let mut right = FastSet::with_capacity(capacity);
 
     // Rotate points clockwise by 45 degrees, scale by √2 and extend edge by 1.
     // This transforms each sensor into an axis aligned bounding box.
@@ -106,6 +107,8 @@ pub fn part2_testable(input: &[Input], size: i32) -> u64 {
     let vertical: Vec<_> = left.intersection(&right).copied().collect();
     let range = 0..=size;
 
+    // Many input files have vertical.len() == 1 and horizontal.len() == 1, which implies exactly
+    // one answer; but this check covers more situations and is not too expensive.
     for &x in &vertical {
         for &y in &horizontal {
             // Rotate intersection point counter-clockwise and scale by 1 / √2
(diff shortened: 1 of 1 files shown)


ebblake:
Admittedly, 50% faster on something that is already near the 1us minimum granularity is a tough sell. I'll understand if you don't choose to apply this.

ebblake:
Huh - my input file has 40 lines, but I found another with only 25. I have to wonder if the set is small enough for an O(n) traversal of a small unsorted Vec is going to be any faster than O(1) FastMap memory management overhead.

src/year2022/day15.rs
```diff
@@ -86,24 +86,23 @@ pub fn part2(input: &[Input]) -> u64 {
 /// Of the entire 4000000 by 4000000 area the missing beacon must be located in the only
 /// square area not covered by a scanner.
 pub fn part2_testable(input: &[Input], size: i32) -> u64 {
-    let mut top = FastSet::new();
-    let mut left = FastSet::new();
-    let mut bottom = FastSet::new();
-    let mut right = FastSet::new();
+    let mut top_bottom: FastMap<_, u8> = FastMap::with_capacity(input.len() * 2);
+    let mut left_right: FastMap<_, u8> = FastMap::with_capacity(input.len() * 2);
```
maneatingape: 80% of the speed up comes from preallocating the set capacity so this PR can be simplified to just:
```
    let mut top = FastSet::with_capacity(50);
    let mut left = FastSet::with_capacity(50);
    let mut bottom = FastSet::with_capacity(50);
    let mut right = FastSet::with_capacity(50);
```

or

```
    let capacity = input.len();
    let mut top = FastSet::with_capacity(capacity);
    let mut left = FastSet::with_capacity(capacity);
    let mut bottom = FastSet::with_capacity(capacity);
    let mut right = FastSet::with_capacity(capacity);
```

src/year2022/day15.rs
```diff
@@ -115,9 +114,12 @@ pub fn part2_testable(input: &[Input], size: i32) -> u64 {
             // As we're mixing overlaps from different boxes there may be some spurious false
             // positives, so double check all points are within the specified area
             // and outside the range of all scanners.
-            if range.contains(&point.x)
-                && range.contains(&point.y)
-                && input.iter().all(|i| i.sensor.manhattan(point) > i.manhattan)
+            // While the example has multiple points to chew through, input files are more likely
```
maneatingape: Valid speedup, but this check takes ~20ns so we can keep the simpler check.

maneatingape (changes requested):

src/year2022/day15.rs
ebblake: Indeed. Patch simplified accordingly. Other arrangements, such as using Vec instead of HashSet to build unordered lists and then do O(n^2) comparisons instead of O(n) intersect, were still in the noise compared to keeping this version as-is. I've updated the PR to match the simplification of the effort.

src/year2022/day15.rs
ebblake: Fair enough. I left the code, but still thought a comment may be helpful to a future reader.

maneatingape (approved):
👍

PR #110, opened 2026-09-09
Sum rows before extrapolating with binomial coefficients

ebblake:
## Description

Since each row represents a polynomial, the summation of the next element of each row will match the next element of the sequence created by summing each row.  But performing 200 rows of 21 multiplications each is slightly slower than performing 200 summations of 21 columns followed by just 21 multiplications.  With this change, it is no longer necessary to save triangle[] for future rows, but rather just compute it on the fly.  Additionally, avoiding input.lines() allows fewer passes over the input bytes.

On my laptop, runtime improves from 31us to 24us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`

### DIFF
--- a/src/year2023/day09.rs
+++ b/src/year2023/day09.rs
@@ -37,6 +37,10 @@
 //! Let `n` be the number of values and `k` the index of each value. The coefficient for each value
 //! is `(n k)` if `k` is even or `-(n k)` if `k` is odd. For part one we then flip the sign of the
 //! sum when `n` is odd.
+//!
+//! Each row of input represents points along a polynomial. Summing all rows together also forms a
+//! sequence of points from a polynomial, and extending the terms of that summation gives the same
+//! results as summing the extension of each row in isolation, but with less overall work.
 use crate::util::parse::*;
 
 type Input = (i64, i64);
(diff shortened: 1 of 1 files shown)


src/year2023/day09.rs
```diff
@@ -37,34 +37,39 @@
...
 
 pub fn parse(input: &str) -> Input {
     // Determine how many numbers are on each row. Assume each row has the same amount.
     let first = input.lines().next().unwrap();
-    let row = first.iter_signed::<i64>().count() as i64;
+    let count = first.iter_signed::<i64>().count() as i64;
+
+    // Sum all rows together before extrapolating terms.
+    let mut terms = vec![0_i64; count as usize];
+    for line in input.lines() {
```
maneatingape: Can collapse into a single loop e.g.

```
    for (i, value) in (0..count).cycle().zip(input.iter_signed::<i64>()) {
        terms[i] += value;
    }
```

This avoid the relatively slow `lines()` method and shaves some more time off.

src/year2023/day09.rs
```diff
@@ -37,34 +37,39 @@
...
-    // Use adjusted binomial coefficients to calculate answers for each row.
     let mut part_one = 0;
     let mut part_two = 0;
 
-    for line in input.lines() {
-        for (k, value) in line.iter_signed::<i64>().enumerate() {
-            part_one += value * triangle[k];
-            part_two += value * triangle[k + 1];
-        }
+    for (k, value) in terms.iter().enumerate() {
```
maneatingape: Can avoid the cast to `i64` with `for (value, k) in terms.iter().zip(0..) {`

maneatingape (changes requested):
This is clever and shaves off some time. I had thought that multiplication and addition were both 1 cycle on all modern processors but apparently there is still some difference.

On the other hand, avoiding the slow `lines()` method gave even higher savings. Using the original code, but removing the nesting and changing to

```
    for (k, value) in (0..count).cycle().zip(input.iter_signed::<i64>()) {
        part_one += value * triangle[k];
        part_two += value * triangle[k + 1];
    }
```

Interestingly this was faster on the original code, than using the same trick on this patch.

ebblake:
> This is clever and shaves off some time. I had thought that multiplication and addition were both 1 cycle on all modern processors but apparently there is still some difference.

x86_64 takes one cycle latency per ADD with up to four running in parallel (in part because the processor is often computing the next address to read/write to in parallel with a value to be written), but 3 cycles latency per IMUL with a pipeline that can still spit out one multiply per clock if data is present.  So an unrolled loop of dependent adds vs. multiplies is about the same (the data dependencies resulting in throughput of one operation per cycle either way), but an algorithm that only adds can often beat an algorithm with multiplies when the compiler/processor can merge independent actions into the same cycle.

maneatingape:
This PR benchmarks at 16,196ns. The same loop unnesting applied to the original benchmarks at 15,652.

Not sure exactly why that's the case but I'll stick with the original and apply the unnesting.

However since it was inspired by your change, happy if you want to make those changes in this PR for the credit.
Alternatively I can close this PR and make them seperately, easy either way.

ebblake:
> This PR benchmarks at 16,196ns. The same loop unnesting applied to the original benchmarks at 15,652.
> 
> Not sure exactly why that's the case but I'll stick with the original and apply the unnesting.
> 
> However since it was inspired by your change, happy if you want to make those changes in this PR for the credit. Alternatively I can close this PR and make them seperately, easy either way.

I'm fine if you prefer applying your own change (it was your idea to use .cycle() after all); and at that point, it diverges enough from my original patch that I don't need the credit - we still get the speedup either way.

PR #111, opened 2026-09-18
Use trapezoid rather than shoelace formula

ebblake:
## Description

The compiler can't see that c[2..c.len()-2] is a fixed-length slice; giving a constant for the endpoint leads to a faster parse.  Also, when all edges are parallel to the axes, the trapezoid formula results in fewer multiplications than the shoelace formula.

On my laptop, this cuts runtime from 20+0.8+0.8 to 15+0.6+0.6 us.

## Type of change

- [X] Performance improvement
- [ ] Bug fix
- [ ] Other

## Checklist

- [X] Tests pass `cargo test`


src/year2023/day18.rs
```diff
@@ -31,7 +31,7 @@ pub fn parse(input: &str) -> Input {
                 b'3' => UP,
                 _ => unreachable!(),
             };
-            let hex = &c[2..c.len() - 2];
+            let hex = &c[2..7];
```
maneatingape: ❤️

src/year2023/day18.rs
```diff
@@ -49,22 +49,26 @@ pub fn part2(input: &Input) -> i64 {
...
+    // Instead of computing the area by the shoelace formula (a determinant per vertex), it is
+    // faster to use the trapezoid formula: sum((y[i]+y[i+1])*(x[i]-x[i+1])). But all vertical
+    // segments have x[i]-x[i+1]=0, so only the horizontal segments, where y[i]+y[i+1]=2y,
+    // actually contribute to the area.
     for &(direction, amount) in moves {
-        let previous = position;
-        position += direction * amount;
-        area += determinant(previous, position);
-        perimeter += amount as i64;
+        let amount = amount as i64;
```
maneatingape: This simplifies to:
```
    for &(direction, amount) in moves {
        half_area += amount * height * direction.x as i64;
        height += amount * direction.y as i64;
        perimeter += amount;
    }
```

maneatingape (changes requested):
The speedup is marginal, but I like the simplification of the code and the alternate version of the Shoelace formula.

BTW changing from `i32` to `i64` in this type
```
type Move = (Point, i64);
```
eliminates a cast.

maneatingape (approved):
👍

